PR #1, opened 2017-08-23
descriptions of plugins

CermakM:
-l, --list-available-tools now print name and description (if
specified)

### DIFF
--- a/py/csmock
+++ b/py/csmock
@@ -40,7 +40,7 @@ patch_rawbuild = csmock_datadir + "/scripts/patch-rawbuild.sh"
 
 cwe_list_file = csmock_datadir + "/cwe-map.csv"
 
-plugin_dir= "@PLUGIN_DIR@"
+plugin_dir= "plugins"
 
 default_cswrap_timeout = 30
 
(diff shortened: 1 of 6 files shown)


py/csmock
```diff
@@ -562,6 +562,14 @@ class PluginManager:
                 continue
             self.try_load(parts[0], path=plugin_dir)
 
+    # Print description of each available plugin in format name [:space:]+ description
+    def print_plugin_descriptions(self):
+        max_indent = max(map(len, self.plug_by_name.keys()))
```
kdudka: ```s/max_indent/max_key_len/``` ?

py/csmock
```diff
@@ -562,6 +562,14 @@ class PluginManager:
...
+        for key, plugin in sorted(self.plug_by_name.items()):
+            desc = getattr(plugin, "description", "")
+            sys.stdout.write("{}{}{}\n".format(key, " "*(max_indent-len(key)+4),
```
kdudka: Please use a named variable (e.g. ```indent_len```?) for ```max_key_len + 4```.  Ideally the ```+ 4``` should be at a single place in the code, so the we do not need to find all occurrences in case we wanted to use ```+ 8``` instead of ```+ 4``` later on.

Please separate binary operators with spaces, e.g.: ```" "*(max_indent-len(key)+4``` --> ```" " * (max_indent  - len(key) + 4```

py/plugins/gcc.py
```diff
@@ -30,7 +30,8 @@ class Plugin:
     def __init__(self):
         self.enabled = False
         self.flags = flags_by_default()
-
+        self.description = "Plugin capturing GCC warnings.\n" \
+                           "Possibly customizable with compilers flags."
```
kdudka: The wording (originally suggested by me IIRC) is slightly misleading.  I would rephrase it as follows:
> Plug-in capturing GCC warnings, optionally with customized compiler flags.

PR #4, opened 2017-10-19
Fix pylint reports

kdudka:
This PR fixes more than 300 issues reported by pylint.  Originally proposed by @CermakM in #3.


kdudka:
Thanks for review!  I have amended the respective commits.

PR #10, opened 2018-10-02
csbuild -g: search parent directories for .git

### LINKED ISSUE #7: csbuild and --git-commit-range don't work for out of source builds
csbuild and --git-commit-range don't work for out of source builds for example which cmake encourages.

So the line: `repo = git.Repo(".")` fails. Either this should look in the parent directories or your provide an option to tell it where to find it.

kdudka:
... even with recent versions of GitPython

Fixes #7

### DIFF
--- a/py/csbuild
+++ b/py/csbuild
@@ -19,6 +19,7 @@
 
 # standard imports
 import argparse
+import inspect
 import os
 import shutil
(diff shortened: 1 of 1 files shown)


cryptomilk:
Tested and works!

PR #11, opened 2018-10-02
doc: clarify the format expected by --git-commit-range

### LINKED ISSUE #8: Improve documentation
The manpage doesn't say in which format GIT_COMMIT_RANGE should be, there should be an example for it.

Also default values should be indicated, e.g.:

```
--added-exit-code ADDED_EXIT_CODE
              exit code to return if there is any defect added in the new version (default: 0)
```

kdudka:
... and document default values of cli options consistently

Fixes #8

### DIFF
--- a/py/csbuild
+++ b/py/csbuild
@@ -266,7 +266,7 @@ def main():
     # optional arguments
     parser.add_argument(
         "-g", "--git-commit-range",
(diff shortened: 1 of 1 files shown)


cryptomilk:
Looks fine for me.

PR #21, opened 2018-10-23
csbuild: use `command -v` instead of `which`

### LINKED ISSUE #20: csbuild fails when run in in a gitlab CI container
csbuild fails with the following error:

```
/usr/share/csbuild/scripts/run-scan.sh: error: /builds/cryptomilk/libssh-mirror/obj-csbuild/CMakeCache.txt contains 'CMAKE_C_COMPILER=/usr/lib64/cscppc/cc' instead of 'CMAKE_C_COMPILER=', which is required to wrap the compiler.  Please run cmake through csbuild to fix this.
```

The path in the cache looks correct, but the expected check seems to be wrong as it is empty.

kdudka:
The external `which` command might not be available on lightweight
installations of Linux distributions.

Suggested-by: Andreas Schneider
Fixes #20

### DIFF
--- a/scripts/run-scan.sh
+++ b/scripts/run-scan.sh
@@ -94,7 +94,7 @@ if test -r CMakeCache.txt; then
             2>/dev/null | cut -c4-)"
         test -n "$abs_path" || continue
         rel_path="$(basename "$abs_path")"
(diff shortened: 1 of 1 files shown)


cryptomilk:
LGTM :-)

PR #29, opened 2021-06-29
scripts/chroot-fixups/openssl-public-header-files.sh: patch `<openssl/evp.h>`

kdudka:
... as proposed in https://github.com/openssl/openssl/pull/15905 in case openssl-3.0 is installed in the buildroot.


sahanaprasad07 (approved):
LGTM

kdudka:
Thanks for review!

PR #32, opened 2021-07-15
improve integration with upstream csmock

kdudka:
... based on the following upstream pull requests:
https://github.com/kdudka/csmock/pull/30 - csmock: do not hard-wire ordering of plug-ins
https://github.com/kdudka/csmock/pull/31 - csmock: handle props.add_repos in try_install()


kdudka:
Oops, I confused two forks with each other.  This wa the right place to open the pull request: https://github.com/aufover/csmock/pull/1

PR #39, opened 2021-09-23
plugins: do not pass `--quiet` to `csgrep` for CBMC, Divine and Symbiotic

lzaoral:
... so that we can spot errors in our converters more easily.  And fail if the conversion fails too.


kdudka:
Merging, thanks!

PR #40, opened 2021-10-01
csmock: assume plug-ins are experimental

kdudka:
... unless explicitly tagged as stable.  We have more experimental plug-ins than stable these days.  If someone forgot to tag a plug-in as experimental, it might have been enabled by mistake.  Assuming that untagged plug-ins are experimental is the safer approach here.

### DIFF
--- a/py/plugins/cbmc.py
+++ b/py/plugins/cbmc.py
@@ -27,7 +27,6 @@ DEFAULT_CBMC_TIMEOUT = 42
 class PluginProps:
     def __init__(self):
         self.description = "Bounded Model Checker for C and C++ programs"
-        self.experimental = True
 
         # hook this plug-in before "gcc" to make ScanProps:enable_csexec() work
         self.pass_before = ["gcc"]
--- a/py/csmock
+++ b/py/csmock
@@ -576,15 +576,22 @@ class PluginManager:
         max_key_len = max(map(len, self.plug_by_name.keys()))
         min_indent_len = 8
         description_indent = max_key_len + min_indent_len
-        for key, plugin in sorted(self.plug_by_name.items()):
-            props = plugin.get_props()
-            desc = getattr(props, "description", "")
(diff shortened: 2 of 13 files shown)


py/plugins/cbmc.py
```diff
@@ -27,7 +27,6 @@
 class PluginProps:
     def __init__(self):
         self.description = "Bounded Model Checker for C and C++ programs"
```
vmihalko, line 29: I miss the dot at the end of the sentence...

vmihalko (approved):
This instructive change looks good to me!
```
$ csmock --list-available-tools
Tools list: 
clang             Source code analysis tool that finds bugs in C, C++, and Objective-C programs.
cppcheck          Static analysis tool for C/C++ code.
gcc               Plugin capturing GCC warnings, optionally with customized compiler flags.
shellcheck        A static analysis tool that gives warnings and suggestions for bash/sh shell scripts.
bandit            [EXPERIMENTAL] A tool designed to find common security issues in Python code.
cbmc              [EXPERIMENTAL] Bounded Model Checker for C and C++ programs
divine            [EXPERIMENTAL] A formal verification tool based on explicit-state model checking.
pylint            [EXPERIMENTAL] Python source code analyzer which looks for programming errors.
smatch            [EXPERIMENTAL] Source code analysis for C, based on sparse.
strace            [EXPERIMENTAL] A dynamic analysis tool that records system calls associated with a running process.
symbiotic         [EXPERIMENTAL] A formal verification tool based on instrumentation, program slicing and KLEE.
valgrind          [EXPERIMENTAL] A dynamic analysis tool for finding memory management bugs in programs.
```

py/plugins/cbmc.py
kdudka: No worries, it is not a sentence anyway.  We might unify such details later on.

kdudka:
Thanks for review!

PR #41, opened 2021-10-04
add --gcc-analyzer-package and --gcc-analyzer-bin flags to gcc plugin

sibeream:


py/plugins/gcc.py
```diff
@@ -91,6 +92,11 @@ def init_parser(self, parser):
             "--gcc-sanitize-undefined", action="store_true",
             help="enable %%check and compile with -fsanitize=undefined")
 
+        parser.add_argument(
+            "--gcc-analyzer-bin", action="store",
+            help="use DTS build of gcc to perform scan"
```
kdudka, line 97: I would say `custom` rather than `DTS` because the option itself is more generic.  I would reword it like this:

> use a custom gcc binary for -fanalyzer

py/plugins/gcc.py
```diff
@@ -158,15 +164,32 @@ def handle_args(self, parser, args, props):
...
+                    return rc
+                verstr = verstr.partition('\n')[0]
+                ver = re.sub("^gcc \(GCC\) ", "", verstr.strip())
+                results.ini_writer.append(prefix, ver)
+                return 0
+
+            return store_gcc_analyzer_version
+
+        props.post_depinst_hooks += \
+            [get_gcc_version_writer("analyzer-version-gcc", analyzer_bin)]
```
kdudka, line 183: This does not seem to work as expected.  It reports the custom gcc version as the system gcc version by mistake.

py/plugins/gcc.py
```diff
@@ -185,7 +208,9 @@ def csgcca_hook(results, mock):
                     props.env["CSGCCA_ADD_OPTS"] = csmock.common.cflags.serialize_flags(args.gcc_analyze_add_flag)
 
                 # record that `gcc -fanalyzer` was used for this scan
-                csmock.common.util.write_toolver_from_rpmlist(results, mock, "gcc", "gcc-analyzer")
+                gcc_version_writer = get_gcc_version_writer("gcc-analyzer", analyzer_bin)
```
kdudka, line 211: The metadata entry should actually be `analyzer-version-gcc-analyzer`.  Please compare the output with the stable version of csmock.

kdudka (changes requested):
I do not think this could really work without updating `props.env["CSGCCA_ANALYZER_BIN"]`.  Minor review comments inline...

py/plugins/gcc.py
sibeream: Reverted to the original code to print gcc version.

py/plugins/gcc.py
```diff
@@ -91,6 +92,11 @@ def init_parser(self, parser):
             "--gcc-sanitize-undefined", action="store_true",
             help="enable %%check and compile with -fsanitize=undefined")
 
+        parser.add_argument(
+            "--gcc-analyzer-bin", action="store",
+            help="use custom build of gcc to perform scan"
+        )
+
```
kdudka, line 99: I think `--gcc-analyzer-bin` should imply `--gcc-analyze` (which enables the `gcc` plug-in).  The new option should be listed right after `--gcc-analyze`, not after `--gcc-sanitize-*`.

py/common/util.py
```diff
@@ -47,6 +47,10 @@ def translate_one(i):
     return cmd_out.lstrip()
 
 
+def write_toolver(results, tool_key, ver):
```
kdudka, line 50: This method should take `ini_writer` as the first parameter because it does not use anything else from `results` (unlike `write_toolver_from_rpmlist()`).

py/plugins/gcc.py
```diff
@@ -16,6 +16,7 @@
 # along with csmock.  If not, see <http://www.gnu.org/licenses/>.
 
 # standard imports
+import re
```
kdudka, line 19: I guess this line should not be added unless the module is used anywhere in `gcc.py`.

py/plugins/gcc.py
```diff
@@ -162,13 +168,17 @@ def handle_args(self, parser, args, props):
...
-                cmd += " | gcc -xc - -c -o /dev/null"
+                cmd += " | %s -xc - -c -o /dev/null" % analyzer_bin
                 cmd += " -fanalyzer -fdiagnostics-path-format=separate-events"
                 if 0 != mock.exec_mockbuild_cmd(cmd):
-                    results.error("`gcc -fanalyzer` does not seem to work, disabling the tool", ec=0)
+                    results.error("`%s -fanalyzer` does not seem to work, "
+                                  "disabling the tool" % analyzer_bin, ec=0)
                     return 0
 
+                props.env["CSGCCA_ANALYZER_BIN"] = analyzer_bin
```
kdudka, line 180: I would export the variable only if `--gcc-analyzer-bin` is actually used.  Otherwise it would unnecessarily pollute the scan log.

sibeream:
> This helped me to make it actually work:
> 
>     1. Create an executable script `/usr/bin/gcc-11` in the buildroot with the following content:
> 
> 
> ```
> #!/bin/bash
> exec /opt/rh/gcc-toolset-11/root/usr/bin/gcc "$@"
> ```
> 
>     1. Create a symlink `/usr/lib64/cswrap/gcc-11` -> `../../bin/cswrap` in the buildroot
> 
>     2. Give `--gcc-analyzer-bin=gcc-11` to csmock.
> 
> 
> If this is the way to go, we need to figure out how to provide the functionality through a user-friendly and flexible enough interface...

The solution which came to me is to inject `/opt/rh/gcc-toolset-11/root/usr/bin/` in PATH variable right after `cswrap` and use `gcc` as analyzer name in this code:
https://github.com/csutils/cscppc/blob/d2c4a2ec48675d8777e4631e08092f80ab3ef58e/src/cswrap-core.c#L396

I.e. `--gcc-analyzer-bin ` and `CSGCCA_ANALYZER_BIN` should provide not the path to analyzer binary, but the path to the directory, containing binary named gcc, which will be used as analyzer.

The limitation created by this solution would be that binary has to be named `gcc`, however since analyzer flags are hardcoded in csgcca in the way they are used in GCC, I would say we already limit options for binary to be GCC or at least to interpret flags the same way GCC doing. Which in my opinion effectively means that we expect analyzer to be GCC.

Also per my understanding in execution branch where analyzer is actually executed, we don't expect the system compiler to be executed in the future, so every process that reached the point to execute analyzer can safely mask all traces of the system compiler in the PATH variable.

PR #43, opened 2021-11-03
Update some ShellCheck CWE mappings

stevegrubb:
During a review of the most common ShellCheck warnings, there were a few
that could better pinpoint the kind of flaw that was being reported.
This patch updates the mapping to be a little more precise.


kdudka:
Merging, thanks!

PR #46, opened 2021-11-09
make-srpm.sh: make cbmc plug-in require cbmc-utils

kdudka:
The cbmc plug-in took care of installing cbmc-utils into buildroot
(mainly to make csexec work).  However its filter_hook() runs on the
host and uses cbmc-convert-output, which is provided by cbmc-utils.
Hence we need cbmc-utils installed also on the host.

Reported-by: Jerry James
Closes: https://github.com/csutils/csmock/pull/46

### DIFF
--- a/make-srpm.sh
+++ b/make-srpm.sh
@@ -134,6 +134,7 @@ This package contains the bandit plug-in for csmock.
 
 %package -n csmock-plugin-cbmc
 Summary: csmock plug-in providing the support for cbmc
+Requires: cbmc-utils
 Requires: csexec
 Requires: csmock-common(python3)
 


lzaoral:
Divine and Symbiotic should have the same issue with their converting scripts, right?

kdudka:
Yes, the divine plug-in seems to use just `csgrep` in its `filter_hook()`.  I will push a follow-up fix for symbiotic, although it would be better to package `symbiotic2cs` separately.  Installing the whole symbiotic package on the host in order to be able to run just the conversion script sounds like overkill.

lzaoral:
I'll probably change it to the same approach as Divine uses because you can't easily add copr repository as BuildRequires.

kdudka:
Both the dependencies are `Requires` in `csmock.spec`.  But feel free to work on further improvements on top of this pull request.

kdudka:
But I see your point.  Fedora packages should not have hard-requirements on packages that are not included in Fedora yet.

kdudka:
I have dropped the symbiotic-related commit from this pull request.  @lzaoral Please open a separate pull request for Symbiotic to run `symbiotic2cs` in the buildroot.

vmihalko:
Thank you @kdudka!

PR #47, opened 2021-11-09
Symbiotic + Divine follow-up to #46

lzaoral:
Do not require any tools that are not available in official Fedora repositories and fix `gclang`'s really annoying warnings about `-disable-llvm-passes` option.


kdudka:
This commit just removes the code that used to run the conversion script.  We should probably document where the code was moved?  Is it now embedded in the `symbiotic` package directly?  If so, shouldn't we wait with the merge until the updated Symbiotic builds are available in Copr?

lzaoral:
Yes, it was moved to `csexec-symbiotic.sh` and the package is already available: https://copr.fedorainfracloud.org/coprs/g/aufover/symbiotic/build/2941177/

PR #48, opened 2021-11-10
Add options to enable parallel run

psimovec:
added argument:
--root-override sets --config-opts=root= for mock
disables yum_cache and root_cache, required for parallelly running other instances of csmock

### DIFF
--- a/py/csmock
+++ b/py/csmock
@@ -189,8 +189,11 @@ class MockWrapper:
     def __init__(self, results, props):
         self.results = results
         self.mock_profile = props.mock_profile
-        self.lock_file = "/tmp/.csmock-%s.lock" % self.mock_profile
-        self.meta_lock_file = "/tmp/.csmock-%s.metalock" % self.mock_profile
+        self.mock_root_override = props.mock_root_override
+        lock = f"/tmp/.csmock-" + (self.mock_profile if not self.mock_root_override else
+                                   self.mock_root_override)
+        self.lock_file = f"{lock}.lock"
+        self.meta_lock_file = f"{lock}.metalock"
         self.pid = os.getpid()
         self.scrub_done = props.skip_mock_init
         self.init_done = props.skip_mock_init
@@ -212,7 +215,12 @@ if test -e \"$lock_file\"; then \n\
 fi \n\
 echo \"$self_pid\" > \"$lock_file\"'" \
             % (MOCK_WAITING_TICK, self.meta_lock_file, self.lock_file, self.pid)
+        advice = False
         while os.system(cmd) != 0:
+            if not advice:
+                self.results.print_with_ts("tip: you can use --root-override=<directory> "
+                      "to run this csmock instance in parallel")
+                advice = True
             f = open(self.lock_file)
             other_pid = ""
             if f is not None:
@@ -240,6 +248,10 @@ echo \"$self_pid\" > \"$lock_file\"'" \
 
         # re-enable verbose output per https://bugzilla.redhat.com/1166609
         self.def_cmd += ["--config-opts=print_main_output=True"]
+        if self.mock_root_override:
+            self.def_cmd += ["--disable-plugin=root_cache",
+                             "--disable-plugin=yum_cache"]
+            self.def_cmd += [f"--config-opts=root={self.mock_root_override}"]
 
         return self
 
@@ -419,6 +431,7 @@ class ScanProps:
         self.base_srpm = None
         self.mock_profile = None
         self.base_mock_profile = None
+        self.mock_root_override = None
         self.any_tool = False
         self.nvr = None
         self.pkg = None
@@ -833,11 +846,16 @@ key event (defaults to 3).")
 
     parser.add_argument(
         "--base-srpm",
-        help="perform a differential scan against the specified base pacakge")
+        help="perform a differential scan against the specified base package")
 
     parser.add_argument(
         "--base-root", dest="base_mock_profile",
         help="mock profile to use for the base scan (use only with --base-srpm)")
+
+    parser.add_argument(
+        "--root-override", dest="mock_root_override",
+        help='override the build root directory for mock (disables yum and root cache)'
+    )
 
     # --skip-patches, --diff-patches, and --shell-cmd are mutually exclusive
     group = parser.add_mutually_exclusive_group()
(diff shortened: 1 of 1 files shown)


kdudka (changes requested):
Let's do just one thing at a time and do it properly.  If I remember correctly, we needed just the `--resultdir` option.  So I would not add the other two options until we have a use case for them.  We already have too many options in csmock.

Please also remove all unrelated changes from commit(s) that add new features or change the behavior, like the reorder/reindent of imports.

And please avoid any non-English comments in public pull requests.

kdudka:
Sorry for the typo.  I meant `--rootdir`, of course.

psimovec:
> Let's do just one thing at a time and do it properly. If I remember correctly, we needed just the `--rootdir` option. So I would not add the other two options until we have a use case for them. We already have too many options in csmock.


That's the problem, I'd love to just keep just --rootdir, but as I have commented I had 0 success with --rootdir. --rootdir just does not work the same way as --config-opts=root=directory as we have expected.

kdudka:
I meant csmock's `--rootdir` option.  It can be implemented by using multiple mock's option underneath, of course.  Let's just not complicate csmock's interface by these implementation details.

As for the locking issue, I believe it could be easily fixed.  `MockWrapper` just needs to use something else as the lock name in case `--rootdir` is used.

psimovec:
> I meant csmock's `--rootdir` option. It can be implemented by using multiple mock's option underneath, of course. Let's just not complicate csmock's interface by these implementation details.
> 
> As for the locking issue, I believe it could be easily fixed. `MockWrapper` just needs to use something else as the lock name in case `--rootdir` is used.

Changing lock name in MockWrapper has no effect. Running a process in parallel causes a quick error:
```
2021-11-11 08:55:53	"/usr/bin/mock" "-r" "default" "--plugin-option=tmpfs:keep_mounted=True" "--config-opts=print_main_output=True" "--rootdir=/var/lib/mock/pes71" "--disable-plugin=root_cache" "--disable-plugin=yum_cache" "--init"
INFO: mock.py version 2.12 starting (python version = 3.9.7, NVR = mock-2.12-1.fc34)...
Start(bootstrap): init plugins
INFO: selinux enabled
Finish(bootstrap): init plugins
Start: init plugins
INFO: selinux enabled
Finish: init plugins
INFO: Signal handler active
Start: run
Start: clean chroot
ERROR: Build root is locked by another process.

!!! 2021-11-11 08:55:54	fatal error: failed to init mock profile (default)
```

Let me point out again, `--config-opts=root=directory` works for me. `--rootdir` is not a simple shortcut for `--config-opts=root=` as it for some reason behaves differently, especially when run in parallel. I think we should use only `--config-opts=root=` argument on mock (but we can call it differently in csmock), and to remove `--rootdir` argument.

kdudka:
OK, got it.  This is a problem with the lock of **mock** itself.  If it works fine with `mock --config-opts=root=directory ...`, let's use that and introduce a new option in csmock for it.  I would just use a different name for the csmock option when it expects a different value than mock's `--rootdir`.  What about `--root-override`?

py/csmock
```diff
@@ -212,6 +216,9 @@ if test -e \"$lock_file\"; then \n\
 fi \n\
 echo \"$self_pid\" > \"$lock_file\"'" \
             % (MOCK_WAITING_TICK, self.meta_lock_file, self.lock_file, self.pid)
+        if os.system(cmd) != 0:
+            print("\ntip: you can use --no-cache --root-override=<directory> "
+                  "to run this csmock instance in parallel")
```
kdudka, line 221: This should be printed to log via `self.results.print_with_ts()` rather than to standard output.

py/csmock
```diff
@@ -874,6 +894,9 @@ key event (defaults to 3).")
         "--version", action=VersionPrinter,
         help="print the version of csmock and exit")
 
+    parser.add_argument("--no-cache", action="store_true",
```
kdudka, line 897: I do not think we need `--no-cache` in csmock's public interface when its only use case is in combination with `--root-override`.  This should be an implementation detail internal to the `MockWrapper` class.

kdudka (changes requested):

py/csmock
kdudka: Also the extra invocation of `cmd` should not be needed to print this message.

py/csmock
```diff
@@ -189,8 +189,10 @@ class MockWrapper:
     def __init__(self, results, props):
         self.results = results
         self.mock_profile = props.mock_profile
-        self.lock_file = "/tmp/.csmock-%s.lock" % self.mock_profile
-        self.meta_lock_file = "/tmp/.csmock-%s.metalock" % self.mock_profile
+        lock = f"/tmp/.csmock-{self.mock_profile}" +\
+            f"-{props.mock_root_override}" if props.mock_root_override else ""
```
kdudka, line 193: The lock name is not optimal.  If we run two instances of csmock like this:
```
csmock -r fedora-34-x86_64 --root-override=xxx ...
csmock -r fedora-35-x86_64 --root-override=xxx ...
```
... they clash with each other.  With `--root-override` specified, we should rather use `self.mock_root_override` on its own as a substitution for `self.mock_profile` in the lock name.  This will also protect the case where someone uses `--root-override=fedora-35-x86_64` as long as system-provided mock profiles use their name for the `root` config option.  This seems to be the case for all profiles in `mock-core-configs` and `csmock-core-configs`.

Please move the `self.mock_root_override = props.mock_root_override` up, so that we do not mix accesses via `self.` and `props.` in the substitutions.

py/csmock
```diff
@@ -212,7 +215,12 @@ if test -e \"$lock_file\"; then \n\
 fi \n\
 echo \"$self_pid\" > \"$lock_file\"'" \
             % (MOCK_WAITING_TICK, self.meta_lock_file, self.lock_file, self.pid)
+        advice = False
         while os.system(cmd) != 0:
+            if not advice:
+                self.results.print_with_ts("\ntip: you can use --root-override=<directory> "
```
kdudka, line 221: The leading `\n` makes the output look inconsistent:
```
>>> 2021-11-19 10:20:55
tip: you can use --root-override=<directory> to run this csmock instance in parallel
>>> 2021-11-19 10:20:55 waiting till .lock (PID 3652) disappears...
>>> 2021-11-19 10:21:55 waiting till .lock (PID 3652) disappears...
```
I propose to drop it.

kdudka (approved):
Looks perfect, merging now.  Thanks!

PR #50, opened 2021-11-18
divine: tweak arguments passed to Divine

lzaoral:
* `--lart stubs` - Transform all undefined functions into stubs, e.g. Divine will fail if a call to such function is reachable.  Previously, Divine even refused to start if any (e.g. unreachable) undefined function was present in the module.
* `-o exit:ignore` - Do not classify non-zero exit code as an error.


kdudka:
Merging, thanks!

PR #53, opened 2021-12-01
symbiotic: set timeout also for sbt-inst and slicer

kdudka:
Otherwise symbiotic may run for a couple of minutes despite we invoke csmock with `--symbiotic-timeout` set to a couple of seconds.

### DIFF
--- a/py/plugins/symbiotic.py
+++ b/py/plugins/symbiotic.py
@@ -94,11 +94,12 @@ class Plugin:
         props.post_depinst_hooks += [create_cap_dir_hook]
 
         # default symbiotic cmd-line
+        timeout = args.symbiotic_timeout
         wrap_cmd_list = [
                 "--skip-ld-linux",
                 "csexec-symbiotic",
                 "-l", SYMBIOTIC_CAPTURE_DIR,
-                "-s", "--prp=memsafety --timeout=%d" % args.symbiotic_timeout]
+                "-s", f"--prp=memsafety --timeout={timeout} --instrumentation-timeout={timeout} --slicer-timeout={timeout}"]
 
         # append custom args if specified
         # FIXME: what about single arguments with whitespaces?


py/plugins/symbiotic.py
```diff
@@ -94,11 +94,12 @@ def create_cap_dir_hook(results, mock):
         props.post_depinst_hooks += [create_cap_dir_hook]
 
         # default symbiotic cmd-line
+        timeout = args.symbiotic_timeout
         wrap_cmd_list = [
                 "--skip-ld-linux",
                 "csexec-symbiotic",
                 "-l", SYMBIOTIC_CAPTURE_DIR,
-                "-s", "--prp=memsafety --timeout=%d" % args.symbiotic_timeout]
+                "-s", f"--prp=memsafety --timeout={timeout} --slicer-timeout={timeout}"]
```
lzaoral, line 102:
```suggestion
                "-s", f"--prp=memsafety --timeout={timeout} --instrumentation-timeout={timeout} --slicer-timeout={timeout}"]
```
Thanks, @kdudka! But we should also set timeout for instrumentation.

lzaoral (changes requested):

py/plugins/symbiotic.py
lzaoral: That option is, however, undocumented so I've created a PR to add it to Symbiotic's help: https://github.com/staticafi/symbiotic/pull/214

kdudka: I thought that `--slicer-timeout` works for both of them.  Will fix it.  Thanks!

lzaoral: No, these are separate:

https://github.com/staticafi/symbiotic/blob/46a794ba5daaa496f550bfc2193f68e1ae9083b4/lib/symbioticpy/symbiotic/transform.py#L329-L334

https://github.com/staticafi/symbiotic/blob/46a794ba5daaa496f550bfc2193f68e1ae9083b4/lib/symbioticpy/symbiotic/transform.py#L495-L499

lzaoral (approved):
LGTM!

kdudka:
Thanks for review!

PR #56, opened 2021-12-07
plugins: avoid `csgrep: Argument list too long` error

kdudka:
... on glob expansion in filter_hook() and make sure we have read
permission on the capture files in case they were created with too
restrictive umask.

### DIFF
--- a/py/plugins/cbmc.py
+++ b/py/plugins/cbmc.py
@@ -119,13 +119,19 @@ class Plugin:
         # transform XML files produced by cbmc into csdiff format
         def filter_hook(results):
             src_dir = results.dbgdir_raw + CBMC_CAPTURE_DIR
-            dst = "%s/cbmc-capture.js" % results.dbgdir_uni
-            cmd = """
+
+            # ensure we have permission to read all capture files
+            results.exec_cmd(['chmod', '-R', '+r', src_dir])
+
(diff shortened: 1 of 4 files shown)


py/plugins/cbmc.py
```diff
@@ -119,13 +119,19 @@ def create_cap_dir_hook(results, mock):
...
                   set -ex
-                  for file in %s/pid-*.out; do
+                  cd '%s'
+                  for file in pid-*.out; do
                       cbmc-convert-output -a < \"$file\" > \"$file.conv\"
                   done
-                  csgrep --mode=json --remove-duplicates '%s'/pid-*.out.conv > '%s'
-                  """ % (src_dir, src_dir, dst)
+                  csgrep --mode=json --remove-duplicates pid-*.out.conv > '%s'
+                  """ % (src_dir, dst)
```
lzaoral, line 135: Could you use `f-strings` for consistency with the rest?

lzaoral (approved):
Thanks, @kdudka! LGTM (except for a small nitpick).

py/plugins/cbmc.py
kdudka: I was not sure whether they work with the multi-line string literals but they do :-)

kdudka:
Thanks for review!  Merging...

PR #57, opened 2021-12-07
csmock: install csexec into chroot from Fedora/EPEL

kdudka:
... rather than copying the binaries from the host.  csexec built for
the target environment may have additional features (e.g. the --argv0
option of ld.so) that are not available in the host environment.


lzaoral (approved):
That's very reasonable, thanks!

PR #59, opened 2021-12-08
divine: pass `-o ignore:notimplemented` to Divine

lzaoral:
Divine will not report calls to undefined functions as an fatal error.

### DIFF
--- a/py/plugins/divine.py
+++ b/py/plugins/divine.py
@@ -95,9 +95,10 @@ class Plugin:
                 "/usr/bin/csexec-divine",
                 "-l", DIVINE_CAPTURE_DIR,
                 # --lart stubs replaces undefined functions with stubs
+                # -o ignore:notimplemented ignores calls to undefined functions
                 # -o ignore:exit ignores non-zero exit codes
-                "-d", "check --lart stubs -o ignore:exit --max-time %d" \
-                      % args.divine_timeout]
+                "-d", "check --lart stubs -o ignore:notimplemented "
+                      f"-o ignore:exit --max-time {args.divine_timeout}"]
 
         # append custom args if specified
         # FIXME: what about single arguments with whitespaces?


py/plugins/divine.py
```diff
@@ -95,9 +95,10 @@ def create_cap_dir_hook(results, mock):
                 "/usr/bin/csexec-divine",
                 "-l", DIVINE_CAPTURE_DIR,
                 # --lart stubs replaces undefined functions with stubs
+                # -o ignore:notimplemented ignores calls to undefined functions
                 # -o ignore:exit ignores non-zero exit codes
-                "-d", "check --lart stubs -o ignore:exit --max-time %d" \
-                      % args.divine_timeout]
+                "-d", "check --lart stubs -o ignore:notimplemented "
+                      f"-o ignore:exit -max-time {args.divine_timeout}"]
```
kdudka, line 101: It seems that one dash from `--max-time` was removed by mistake, which prevented divine from starting:
```
ERROR: could not parse command
synopsis: /usr/bin/divine check [options] {file} {vector}*
command:  /usr/bin/divine check --lart stubs -o ignore:notimplemented -o ignore:exit -max-time 150 --stdin /tmp/divine-stdinLKBC --capture ././minigzip --threads 2 --report-filename /build
matched:  --lart stubs             as --lart {vector}
matched:  -o ignore:notimplemented as -o{vector}
matched:  -o ignore:exit           as -o{vector}
file -max-time does not exist
failed to match '-max-time'
```

kdudka (changes requested):

py/plugins/divine.py
lzaoral: Good catch! Thanks!

kdudka (approved):
Thanks for the update!  Merging...

PR #61, opened 2022-02-10
cbmc: do not convert the outputs on host

vmihalko:
This change is similar to https://github.com/csutils/csmock/pull/47.
I move the conversion into cbmc-utils/csexec-cbmc.sh: https://github.com/aufover/cbmc-utils/commit/f7b4f2ec40d8c2ed2179de83a936710911a21899.

### DIFF
--- a/py/plugins/cbmc.py
+++ b/py/plugins/cbmc.py
@@ -125,15 +125,8 @@ class Plugin:
 
             # `cd` first to avoid `csgrep: Argument list too long` error on glob expansion
             dst = f"{results.dbgdir_uni}/cbmc-capture.js"
-            cmd = f"""
-                  set -ex
-                  cd '{src_dir}'
-                  shopt -s nullglob
-                  for file in pid-*.out; do
-                      cbmc-convert-output -a < \"$file\" > \"$file.conv\"
-                  done
-                  touch empty.err
-                  csgrep --mode=json --remove-duplicates empty.err pid-*.out.conv > '{dst}'
-                  """
+            
+            cmd = f"csgrep --mode=json --remove-duplicates {src_dir}/pid-*.conv > {dst}"
+            
             return results.exec_cmd(cmd, shell=True)
         props.post_process_hooks += [filter_hook]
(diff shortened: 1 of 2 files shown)


py/plugins/cbmc.py
```diff
@@ -125,15 +125,8 @@ def filter_hook(results):
...
-                  cd '{src_dir}'
-                  shopt -s nullglob
-                  for file in pid-*.out; do
-                      cbmc-convert-output -a < \"$file\" > \"$file.conv\"
-                  done
-                  touch empty.err
-                  csgrep --mode=json --remove-duplicates empty.err pid-*.out.conv > '{dst}'
-                  """
+            
+            cmd = f"csgrep --mode=json --remove-duplicates {src_dir}/pid-*.conv > {dst}"
```
kdudka, line 129: I would prefer to write it as:
```py
 cmd = f"cd '{src_dir}' && touch empty.conv && csgrep --mode=json --remove-duplicates *.conv > '{dst}'"
```

Otherwise we would reintroduce the issue that 5c31ee11547cb3485b4c395db4fb270ea4f7292c tried to avoid (the `touch empty.conv` part was introduced in d47d9b3b2ded03d25f2e002d049c89a0eb3b9978).

kdudka (changes requested):
Thank you for working on it!  Please also remove the dependency on `cbmc-utils` in spec file (embedded in `make-srpm.sh`).

vmihalko:
I'm sorry - this PR completely slipped my mind.

py/plugins/cbmc.py
vmihalko: Resolved in https://github.com/csutils/csmock/pull/61/commits/ef501ee23ebb0d9e241a8f97000405f34e84a8b5.

kdudka:
No worries.  Better later than never :-)

Please also remove the dependency on `cbmc-utils` in spec file (embedded in `make-srpm.sh`).

kdudka:
Merging, thanks!

PR #63, opened 2022-04-27
Align CPPCHECK with other tools

stevegrubb:
A review of some findings across multiple tools is showing that
CPPCHECK uninitvar is using CWE-456 while others use CWE-457.
Both are very similar, but 457 seems more appropriate given that
is how others mark the same finding.


kdudka:
Merging, thanks!

PR #64, opened 2022-05-23
cwe-map.csv: assign specific CWE's to some SHELLCHECK warnings

stevegrubb:


cwe-map.csv
```diff
@@ -391,49 +392,49 @@
 "SHELLCHECK_WARNING","warning[SC2038]","CWE-398"
 "SHELLCHECK_WARNING","warning[SC2039]","CWE-475"
 "SHELLCHECK_WARNING","warning[SC2041]","CWE-398"
-"SHELLCHECK_WARNING","warning[SC2043]","CWE-398"
+"SHELLCHECK_WARNING","warning[SC2043]","CWE-606"
```
kdudka, line 395: As I understand it, `CWE-606` is about excessive looping whereas `SC2043` is about a loop that is always executed only once.  Would not `CWE-561` or `CWE-1164` be more appropriate for this?

kdudka (commented):
Looks good.  One question inline...

cwe-map.csv
stevegrubb: 561 is deadcode. This gets executed, so that's not a good fit. 1164 is similar to dead code where it's removal won't affect the program. In this case, SC2043 is about a bad conditional check. 606, Unchecked Input for Loop Condition, seems like the closest fit because it is about looping and not properly checked inputs. It does mention a consequence being excessive looping, but an improper check could also be no looping. If you like, I can run this by the CWE committee to see if there is a better fit.

stevegrubb:
OK, I see what you mean by removing the code & 1164. I'm looking into this more.

stevegrubb:
I'm going to withdraw this patch. I will remove the 606 suggestion and resubmit. I'll come back with something for SC2043 later.

kdudka:
There is no need to resubmit.  I will just merge this without the `SC2043` change for now.  Thanks for the contribution!

PR #67, opened 2022-06-03
Update GCC CWE mappings

stevegrubb:
This update adds most of the common GCC warnings with their CWE mapping.
With this addition, this CSV file covers more than 95% of what you find
in large scale scanning.


cwe-map.csv
```diff
@@ -31,35 +31,118 @@
...
+"COMPILER_WARNING","warning[-Wpointer-arith]","CWE-468"
 "COMPILER_WARNING","warning[-Wpointer-compare]","CWE-569"
+"COMPILER_WARNING","warning[-Wpointer-sign]","CWE-681"
+"COMPILER_WARNING","warning[-Wpointer-to-int-cast]","CWE-704"
+"COMPILER_WARNING","warning[-Wredundant-decls]","CWE-1041"
+"COMPILER_WARNING","warning[-Wregister]","CWE-758"
+"COMPILER_WARNING","warning[-Wreorder]","CWE-665"
+"COMPILER_WARNING","warning[-Wrestrict]","CWE-9001"
 "COMPILER_WARNING","warning[-Wreturn-local-addr]","CWE-562"
+"COMPILER_WARNING","warning[-Wreturn-type]","CWE-393"
```
kdudka, line 109: `CWE-393` seems accurate.  Should we use it also for the `MISSING_RETURN` checker of Coverity, which overlaps with this?

cwe-map.csv
```diff
@@ -31,35 +31,118 @@
...
+"COMPILER_WARNING","warning[-Warray-parameter]","CWE-1228"
+"COMPILER_WARNING","warning[-Wattribute-warning]","CWE-9001"
+"COMPILER_WARNING","warning[-Wattributes]","CWE-9001"
+"COMPILER_WARNING","warning[-Wbad-function-cast]","CWE-704"
 "COMPILER_WARNING","warning[-Wbool-compare]","CWE-569"
+"COMPILER_WARNING","warning[-Wbool-operation]","CWE-480"
 "COMPILER_WARNING","warning[-Wbuiltin-declaration-mismatch]","CWE-686"
+"COMPILER_WARNING","warning[-Wcast-function-type]","CWE-843"
+"COMPILER_WARNING","warning[-Wcast-qual]","CWE-704"
+"COMPILER_WARNING","warning[-Wchar-subscripts]","CWE-465"
```
kdudka, line 48: What about CWE-823, which is more specific?

cwe-map.csv
```diff
@@ -31,35 +31,118 @@
...
+"COMPILER_WARNING","warning[-Wexpansion-to-defined]","CWE-9001"
+"COMPILER_WARNING","warning[-Wfloat-conversion]","CWE-197"
+"COMPILER_WARNING","warning[-Wfloat-equal]","CWE-1077"
+"COMPILER_WARNING","warning[-Wformat]","CWE-686"
 "COMPILER_WARNING","warning[-Wformat-extra-args]","CWE-685"
+"COMPILER_WARNING","warning[-Wformat-nonliteral]","CWE-134"
+"COMPILER_WARNING","warning[-Wformat-overflow]","CWE-676"
+"COMPILER_WARNING","warning[-Wformat-security]","CWE-134"
 "COMPILER_WARNING","warning[-Wfree-nonheap-object]","CWE-590"
+"COMPILER_WARNING","warning[-Wignored-qualifiers]","CWE-704"
```
kdudka, line 71: What about `CWE-1164` for this?

cwe-map.csv
```diff
@@ -31,35 +31,118 @@
...
+"COMPILER_WARNING","warning[-Wdeprecated]","CWE-477"
+"COMPILER_WARNING","warning[-Wdeprecated-copy-dtor]","CWE-477"
+"COMPILER_WARNING","warning[-Wdeprecated-copy]","CWE-477"
 "COMPILER_WARNING","warning[-Wdeprecated-declarations]","CWE-477"
+"COMPILER_WARNING","warning[-Wdiscarded-qualifiers]","CWE-704"
+"COMPILER_WARNING","warning[-Wduplicated-branches]","CWE-670"
+"COMPILER_WARNING","warning[-Wduplicated-cond]","CWE-670"
 "COMPILER_WARNING","warning[-Wempty-body]","CWE-1071"
+"COMPILER_WARNING","warning[-Wenum-conversion]","CWE-192"
+"COMPILER_WARNING","warning[-Wexpansion-to-defined]","CWE-9001"
```
kdudka, line 62: Could we assign `CWE-758` to this?

cwe-map.csv
```diff
@@ -31,35 +31,118 @@
...
+"COMPILER_WARNING","warning[-Woverflow]","CWE-681"
+"COMPILER_WARNING","warning[-Wparentheses]","CWE-569"
+"COMPILER_WARNING","warning[-Wpointer-arith]","CWE-468"
 "COMPILER_WARNING","warning[-Wpointer-compare]","CWE-569"
+"COMPILER_WARNING","warning[-Wpointer-sign]","CWE-681"
+"COMPILER_WARNING","warning[-Wpointer-to-int-cast]","CWE-704"
+"COMPILER_WARNING","warning[-Wredundant-decls]","CWE-1041"
+"COMPILER_WARNING","warning[-Wregister]","CWE-758"
+"COMPILER_WARNING","warning[-Wreorder]","CWE-665"
+"COMPILER_WARNING","warning[-Wrestrict]","CWE-9001"
```
kdudka, line 107: Could we assign `CWE-758` to this?

cwe-map.csv
```diff
@@ -31,35 +31,118 @@
...
+"COMPILER_WARNING","warning[-Wshift-overflow]","CWE-682"
+"COMPILER_WARNING","warning[-Wsign-compare]","CWE-195"
+"COMPILER_WARNING","warning[-Wsign-conversion]","CWE-681"
+"COMPILER_WARNING","warning[-Wsign-promo]","CWE-196"
+"COMPILER_WARNING","warning[-Wsizeof-array-div]","CWE-682"
+"COMPILER_WARNING","warning[-Wsizeof-pointer-div]","CWE-682"
+"COMPILER_WARNING","warning[-Wsizeof-pointer-memaccess]","CWE-805"
+"COMPILER_WARNING","warning[-Wstack-protector]","CWE-691"
+"COMPILER_WARNING","warning[-Wstack-usage]","CWE-770"
+"COMPILER_WARNING","warning[-Wstrict-aliasing]","CWE-9001"
```
kdudka, line 121: Could we assign `CWE-758` to this?

kdudka (commented):
Looks great.  A few minor suggestions inline...

cwe-map.csv
stevegrubb: It appears that both are correct and related. 393->684->710. If you want to, we can use 393 for Coverity since it's more specific.

cwe-map.csv
stevegrubb: If you look at the taxonomy mappings for 704, there is a Cert-C EXP05-C which says do not cast away const qualifiers. (This is what you would need to do to suppress the warning.) So, this is an indication that the authors intended 704 to encompass any type conversion problems.

cwe-map.csv
stevegrubb: ok

cwe-map.csv
stevegrubb: ok

cwe-map.csv
stevegrubb: This is the kind of warning that I wanted to lump under 9001, Low level non-security issues. This can be a bug for sure. But it's more likely to be a bad calculation than something exploitable. 758 mentions relying on undefined behavior. I don't think this is **relying** on it.  Rather, this is probably some kind of expression issue which can cause the wrong answer.

kdudka: OK, let's keep `CWE-9001` for this.

cwe-map.csv
stevegrubb: It would be more satisfying to have every GCC warning fit nicely into a CWE. But sometimes close is all you can do. 704 is about type conversion or casting problems - they do not enumerate all the permutations of that like buffer overflows. Some of the other items in the taxonomy mapping mention, "access a variable through a pointer of an incompatible type" which would apply. 1164 makes the distinction that removal of the code would make no difference. I don't think that means removal of "const". I think 704 is about as close as we can get.

stevegrubb: In my testing, if you have
const char *foo(void);
char *bar = foo();
you get the warning. But if you have
char *bar = (char *)foo();
you do not. This is why I mentioned Cert-C EXP05-C. To make it go away, you have to add a cast. 1164 can be thought of as dead code. Meaning whole lines or functions can be removed with no effect.

kdudka: @stevegrubb Which warning exactly do you get?

The code above gives me `-Wdiscarded-qualifiers` whereas I was proposing `CWE-1164` to be assigned to `-Wignored-qualifiers`.  They are two different warnings describing two different programming mistakes.

stevegrubb: I guess another option is to drop this mapping altogether. It is not exploitable - it's a diagnostic.

stevegrubb:
OK, the merge request was updated to map Wignored-qualifiers to 1164.

kdudka:
Thanks for the update!  Merging now, will deploy it next week hopefully.

PR #72, opened 2022-08-08
Add Shell linter - Differential-ShellCheck

jamacku:
Differential ShellCheck is a GitHub action that performs differential ShellCheck scans on files changed via PR and reports results directly in PR.

Since this repository contains some shell scripts I think you might find it useful.

Documentation is available at: [@redhat-plumbers-in-action/differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme). Let me know If you are missing some feature or setting.

/cc @kdudka

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+---
+
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+    reviewers:
+      - lzaoral
--- a/.github/workflows/differential-shellcheck.yml
+++ b/.github/workflows/differential-shellcheck.yml
@@ -1,0 +1,28 @@
+---
+
+name: Differential ShellCheck
+on:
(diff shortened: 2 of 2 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+---
+
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+    labels:
```
lzaoral, line 9: Could you please remove these? Dependabot adds that automatically so that were consistent with the rest of csutils projects.

lzaoral (approved):
LGTM! A minor nitpick inline.

.github/dependabot.yml
jamacku: sure :+1:

jamacku: Done

kdudka (approved):
Looks good.  Thanks!

PR #74, opened 2022-10-05
make-srpm.sh: drop leftovers from py2 compatibility

kdudka:


py/csbuild
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python3
```
lzaoral, line 1: I would use this instead:
```
#!/usr/bin/env python3
```
It is more portable and works when you're using python in `venv`.  Also, `rpmbuild` will still mangles this to `#!/usr/bin/python3` so there should be no changes in behaviour from the Fedora point of view.

py/csmock
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python3
```
lzaoral, line 1: Ditto.

lzaoral (changes requested):
Small nitpick inline.

py/csbuild
kdudka: I was not sure what the best practice is, so I looked into `/usr/bin/dnf` and saw `#!/usr/bin/python3`.  I will change it as you suggest....

lzaoral: Yeah, it will always look like that in Fedora packages due to shebang mangling.  The recommended practice is mentioned here: https://docs.python.org/3/using/unix.html#miscellaneous

lzaoral (approved):
Thank you, @kdudka! LGTM.

kdudka:
Thanks for review!

PR #76, opened 2022-10-14
csmock: treat --root argument more like mock

dsugar100:
In mock the --root argument can be full path OR relative to /etc/mock, man page states:

       -r CONFIG, --root=CONFIG
              Uses  specified  chroot  configuration  as defined in ~/.config/mock/<CONFIG>.cfg or /etc/mock/<CONFIG>.cfg.  Optionally if CONFIG
              ends in '.cfg', it is interpreted as full path  to  config  file.  If  none  specified,  uses  the  chroot  config  linked  to  by
              /etc/mock/default.cfg.

This change makes csmock behave more like mock for the --root command line argument

Signed-off-by: Dave Sugar <dsugar100@gmail.com>


py/csmock
```diff
@@ -206,6 +207,9 @@ class MockWrapper:
         self.def_cmd = None
 
     def __enter__(self):
+        lock_dir = pathlib.Path(self.meta_lock_file).parent.absolute ()
```
kdudka, line 210: Would it make sense to replace all `/` with `_` in the lock name to avoid the need to create the directories?

@lzaoral What do you think?

kdudka (commented):
Thank you for opening the merge request.  It makes perfect sense to me.  One minor question inline...

py/csmock
lzaoral: I prefer @kdudka's suggestion.  Btw, @dsugar100, could you please remove the whitespaces between a function name and its argument list? Thank you!

dsugar100: Yes, this is much cleaner.  Let me know if anything else is needed.

kdudka:
@dsugar100 Thanks for the update!  Merging...

PR #77, opened 2022-10-27
Fix compatibility with the removal of distutils module in Python 3.12

lzaoral:
Related: https://lists.fedoraproject.org/archives/list/python-devel@lists.fedoraproject.org/message/N6ITYHLRWIDNYNXGPYG2ZHF3ZLQWZN7L/

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (C) 2014 Red Hat, Inc.
+# Copyright (C) 2014-2022 Red Hat, Inc.
 #
 # This file is part of csmock.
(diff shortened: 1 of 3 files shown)


kdudka (approved):
Looks perfect.  Thanks!

lzaoral:
Note to self: change `FindPython 3` to `FindPython3` in the commit message and remove the `NOT Python3_FOUND` check since it will always evaluate to false because we called `find_package` with a `REQUIRED` flag.

lzaoral:
Thanks for the review, @kdudka! Merging...

PR #79, opened 2022-11-03
gcc: make sure gcc is installed in the chroot

kdudka:
... to avoid a version hook failure in case gcc is not a build dependency of the scanned artifact:
```
!!! 2022-11-03 15:00:19	error: tool "gcc" does not seem to be installed in build root

!!! 2022-11-03 15:00:19	error: post-depinst hook failed
```

Closes: https://github.com/csutils/csmock/pull/79


kdudka:
@lzaoral That was fast :-) Thanks for review!

PR #82, opened 2022-12-08
chroot-fixups/openssl-public-header-files.sh: remove embedded patch

kdudka:
... of `/usr/include/openssl/evp.h`

The patch partially fails to apply in up2date buildroots and we do not have manpower to maintain it downstream forever.  The corresponding upstream pull request has been staled for more than a year: https://github.com/openssl/openssl/pull/15905


kdudka:
@beldmit For you information, the downstream patch of `<openssl/evp.h>` is being dropped.  The `sed` command to enforce `-DDEBUG_UNUSED` stays.

PR #85, opened 2023-01-17
Improve shell script detection

jamacku:


scripts/run-shellcheck.sh
```diff
@@ -2,8 +2,8 @@
 export LC_ALL="C"
 
 find "$@" -type f -print0 \
-    | xargs -0 sh -c 'for i in "$@"; do { printf "%s\n" "$i" | grep "\\.sh\$" \
-        || head -n1 "$i" | grep --text -Ev "(atf-|tcl|wi)sh\$" | grep --text "^#!.*sh\$"
+    | xargs -0 sh -c 'for i in "$@"; do { printf "%s\n" "$i" | grep -E "^.*\\.(ash|bash|bats|dash|ksh|sh)$" \
```
jamacku, line 5: @kdudka Isn't there missing `\`?

```suggestion
    | xargs -0 sh -c 'for i in "$@"; do { printf "%s\\n" "$i" | grep -E "^.*\\.(ash|bash|bats|dash|ksh|sh)$" \
```

kdudka: As discussed in person, both the variants seem to work equally but I will add the backslash for clarity.

kdudka (approved):
Merging, thanks for the contribution!  I will open another pull request to make this code easier to maintain...

.github/workflows/differential-shellcheck.yml
```diff
@@ -22,7 +23,15 @@ jobs:
...
           fetch-depth: 0
 
-      - name: Differential ShellCheck
+      - id: ShellCheck
+        name: Differential ShellCheck
         uses: redhat-plumbers-in-action/differential-shellcheck@v4
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
+
+      - if: ${{ always() }}
```
lzaoral, line 32: Isn't this a bit redundant? 🤔

lzaoral: Never mind, I will run even when the previous step fails, right?

jamacku: correct :+1:

PR #86, opened 2023-01-20
make the `run-shellcheck.sh` script easier to maintain

kdudka:


jamacku (approved):
LGTM, looks much better, thanks. It could also be extended with other matches in future.

kdudka:
@jamacku Thanks for review!

PR #87, opened 2023-01-27
Implement output capture for sanitizers

lzaoral:


lzaoral:
This PR does not implement filtering and conversion of the results. Those depend on https://github.com/csutils/csdiff/issues/108.

py/plugins/gcc.py
```diff
@@ -30,6 +30,7 @@
 CSMOCK_GCC_WRAPPER_TEMPLATE = '#!/bin/bash\n' \
                               'exec %s "$@"'
 
+SANITIZER_CAPTURE_DIR = "/builddir/sanitizer-capture"
```
kdudka, line 33: Should we rather use `/builddir/gcc-sanitize-capture` to avoid clash with other plug-ins that can be written later on?

py/plugins/gcc.py
```diff
@@ -169,8 +186,27 @@ def handle_args(self, parser, args, props):
...
+            if "valgrind" in props.install_pkgs:
+                parser.error("GCC sanitizers are not compatible with valgrind")
+
+            self.flags.append_flags(['-g', '-fno-omit-frame-pointer',
+                                     '-fsanitize-recover=all'])
+
+            # sanitizers are not compatible with FORTIFY_SOURCE
+            # https://github.com/google/sanitizers/issues/247#issuecomment-1283500316
+            self.flags.append_flags(["-Wp,-U_FORTIFY_SOURCE",
+                                     "-fplugin=annobin",
```
kdudka, line 199: Should we handle the case where `annobin` is not installed in the chroot?

kdudka (approved):

py/plugins/gcc.py
lzaoral: It should not be necessary:
```console
$ dnf repoquery --whatrequires gcc-plugin-annobin
Last metadata expiration check: 0:08:56 ago on Po 6. února 2023, 13:35:29.
redhat-rpm-config-0:228-1.fc37.noarch
```

py/plugins/gcc.py
lzaoral: Good idea! Addressed in 2e410f2.

py/plugins/gcc.py
kdudka: Fair enough.  I think there is no guarantee that `annobin` will be in any buildroot but we can defer the solution until this becomes a real problem.  Could you please put there a comment describing our expectations about annobin's presence in the buildroot?

lzaoral: Addressed in eec74d55dde6285c7aa6fa8ab322414920223974.

lzaoral:
@kdudka 10dabe3a082f1b93abd92a49aca68582fdf3d161 should enable ASAN analysis of more packages.

py/plugins/gcc.py
```diff
@@ -139,11 +150,32 @@ def handle_args(self, parser, args, props):
...
+                    return rv
+
+                # FIXME: hack
+                extra_env = {"ASAN_OPTIONS": props.env["ASAN_OPTIONS"] + ",verify_asan_link_order=0"}
+                ec = mock.exec_rpmbuild_bi(props, extra_env=extra_env)
+                if 0 != ec:
+                    results.error(f"ASAN plug-in: %install or %check failed with exit code {rv}")
+                return ec
+
+            props.run_check = False
```
kdudka, line 176: Although the csmock plug-in API is undocumented, the plug-ins are expected to either set these flags, or leave them unchanged.  Explicitly clearing the `props.run_check` flag may silently break other plug-ins enabled at the same time.

lzaoral: Ok, I'll rework it. However, leaving it enabled won't work since the code was built with ASAN. One solution would be to preload ASAN for everything every time, but that is far from ideal.

kdudka: If a combination of plug-ins or their options does not work together, I think it is better to fail, rather than silently paralyzing other plug-ins.  We already have a few `assert` commands in the plug-ins because of this.

lzaoral: Addressed in eda1cf97596e67ecd0f9285dd42834cc0491dc53.

kdudka:
Thanks for the update!  I think this is ready to be merged.

lzaoral:
Thanks for the review!

PR #99, opened 2023-02-14
copy debug files from chroot

kdudka:
... even when a terminating failure is already being handled


lzaoral (approved):
LGTM!

kdudka:
@lzaoral Thanks for review!

PR #100, opened 2023-03-17
gitleaks: implement --gitleaks-rate-limit defaulting to 1024

kdudka:
Gitleaks reports too many warnings on certain packages, which wastes resources for no good reason.  Implement warning rate limiting in the filter hook to make the further processing of results less expensive.

Resolves: https://github.com/csutils/csmock/issues/98
Requires: https://github.com/csutils/csdiff/pull/119
Closes: https://github.com/csutils/csmock/pull/100

Issue #98, opened 2023-02-08
gitleaks: the plug-in should not report 636404 warnings in `dotnet6.0`


kdudka:
After enabling gitleaks by default we hit performance and resource wasting issues with e.g. `dotnet6.0-6.0.114-1.el8_7` where gitleaks report too many warnings, mostly in log files produced during the build:
```
     12 GITLEAKS_WARNING                                    warning[aws-access-token]
 636404 GITLEAKS_WARNING                                    warning[generic-api-key]
     28 GITLEAKS_WARNING                                    warning[private-key]
      2 GITLEAKS_WARNING                                    warning[sumologic-access-id]
```

kdudka:
Similar problem with `python3.11-cryptography-37.0.2-2.el9`:
```
      2	BAD_FREE                                        	incorrect_free
     35	COMPILER_WARNING                                	warning[-Wsign-conversion]
      2	COMPILER_WARNING                                	warning[-Wunused-function]
  49906	GITLEAKS_WARNING                                	warning[generic-api-key]
     73	GITLEAKS_WARNING                                	warning[private-key]
```

kdudka:
... and `nss-3.79.0-11.el8_7`:
```
 156836 GITLEAKS_WARNING                                        warning[generic-api-key]
      3 GITLEAKS_WARNING                                        warning[private-key]
```

PR #104, opened 2023-04-05
gitleaks: install the plug-in by default

kdudka:
... and enable it with `--all-tools`.  The plug-in has been enabled in production OpenScanHub since November 2022 and did not cause any major problems (except issue #98 which was fixed via pull request #100).

Closes: https://github.com/csutils/csmock/pull/104


lzaoral (approved):
LGTM!

kdudka:
@lzaoral Thanks for review!

PR #111, opened 2023-07-11
Add a few more ShellCheck CWE mappings

stevegrubb:
After analyzing a large scan, I found some new common ShellCheck warnings that are unmapped. This gives them a mapping so they can be analyzed better.


kdudka (approved):
Merging, thanks!

PR #113, opened 2023-07-21
csmock: be less verbose while extracting source tarballs

kdudka:
Some source tarballs contain a huge number of files and the verbose output of tar unnecessarily pollutes `scan.log`.


lzaoral (approved):
LGTM!

PR #114, opened 2023-07-27
consume csgrep changes to support a new option: limit-msg-len (OSH-67)

lbossis:
First cut

### DIFF
--- a/make-srpm.sh
+++ b/make-srpm.sh
@@ -124,7 +124,7 @@ Tool for plugging static analyzers into the build process, free of mock.
 
 %package -n csmock-common
 Summary: Core of csmock (a mock wrapper for Static Analysis tools)
-Requires: csdiff > 3.0.2
+Requires: csdiff > 3.0.3
 Requires: csgcca
 Requires: cswrap
 Requires: mock
--- a/py/plugins/gitleaks.py
+++ b/py/plugins/gitleaks.py
@@ -34,7 +34,7 @@ GITLEAKS_OUTPUT = "/builddir/gitleaks-capture.sarif"
 
 GITLEAKS_LOG = "/builddir/gitleaks-capture.log"
 
-FILTER_CMD = "csgrep '%s' --mode=json --warning-rate-limit=%i > '%s'"
+FILTER_CMD = "csgrep '%s' --mode=json --warning-rate-limit=%i --limit-msg-len=%i > '%s'"
 
 
 class PluginProps:
@@ -71,6 +71,10 @@ class Plugin:
         parser.add_argument(
             "--gitleaks-rate-limit", type=int, default=1024,
             help="drop warnings if their count exceeds the specified limit")
+
+        parser.add_argument(
+            "--gitleaks-limit-msg-len", type=int, default=512,
+            help="trim message if it exceeds max message length")
 
         parser.add_argument(
             "--gitleaks-refresh", action="store_true",
@@ -145,7 +149,7 @@ class Plugin:
         def filter_hook(results):
             src = results.dbgdir_raw + GITLEAKS_OUTPUT
             dst = "%s/gitleaks-capture.js" % results.dbgdir_uni
-            cmd = FILTER_CMD % (src, args.gitleaks_rate_limit, dst)
+            cmd = FILTER_CMD % (src, args.gitleaks_rate_limit, args.gitleaks_limit_msg_len, dst)
             return results.exec_cmd(cmd, shell=True)
 
         props.post_process_hooks += [filter_hook]


kdudka:
@lbossis Thanks!  I have not tried it myself but the code change looks good.  Some minor remarks:
- The option needs to start with the `gitleaks` prefix because it is specific the the `gitleaks` plug-in: `--gitleaks-limit-msg-len`
- I would use a higher default value (say 512) for the new option.
- The commit summary should describe the changes from csmock perspective, something like:
```
gitleaks: implement --gitleaks-limit-msg-len defaulting to 512
```
Implementation details can go to commit description in needed.
- We need to make `csmock-plugin-gitleaks` depend on the updated version of `csdiff`:
```diff
--- a/make-srpm.sh
+++ b/make-srpm.sh
@@ -175,6 +175,7 @@ This package contains the divine plug-in for csmock.
 %package -n csmock-plugin-gitleaks
 Summary: experimental csmock plug-in
 Requires: csmock-common
+Requires: csdiff > 3.0.3

 %description -n csmock-plugin-gitleaks
 This package contains the gitleaks plug-in for csmock.
```
Without the dependency, the scans would fail on the missing `csgrep` option if someone forgot to update the `csdiff` package.

py/plugins/gitleaks.py
```diff
@@ -34,8 +34,7 @@
 
 GITLEAKS_LOG = "/builddir/gitleaks-capture.log"
 
-FILTER_CMD = "csgrep '%s' --mode=json --warning-rate-limit=%i > '%s'"
-
+FILTER_CMD = "csgrep '%s' --mode=json --warning-rate-limit=%i --gitleaks-limit-msg-len=%i > '%s'"
```
lzaoral, line 37: The `csdiff` option introduced by https://github.com/csutils/csdiff/pull/134 is called `--limit-msg-len` so this code won't work.  I haven't tested that locally but the following patch should fix it:
```suggestion
FILTER_CMD = "csgrep '%s' --mode=json --warning-rate-limit=%i --limit-msg-len=%i > '%s'"
```

lzaoral (changes requested):

lbossis:
Removed prefix gitleaks- from FILTER_CMD which calls csgrep. Looks like it works! I executed the following command
csmock -t cppcheck,clang,gitleaks --gitleaks-limit-msg-len=24 nss-util-3.90.0-1.el7_9.src.rpm
the tail of output is found at http://pastebin.test.redhat.com/1106663
Please, do your testing and let's merge this branch limit-amount-of-results-produced-by-single-scan--OSH-67 to main.

make-srpm.sh
```diff
@@ -124,7 +124,7 @@ Tool for plugging static analyzers into the build process, free of mock.
 
 %package -n csmock-common
 Summary: Core of csmock (a mock wrapper for Static Analysis tools)
-Requires: csdiff > 3.0.2
+Requires: csdiff > 3.0.3
```
kdudka, line 127: This is in fact a dependency of `csmock-plugin-gitleaks` but bumping the `csmock-common` requirement should be harmless, too.

kdudka (changes requested):
@lbossis Thanks for the update! Now you need to squash all the fix-up commits and rewrite the commit message as described in https://github.com/csutils/csmock/pull/114#issuecomment-1653170888. This way we end up with a clean git history where each snapshot can be built and tested. This is crucial for using git bisect later on.

lbossis:
Hi Lukas,

Currently this piece of code looks like:

GITLEAKS_LOG = "/builddir/gitleaks-capture.log"

FILTER_CMD = "csgrep '%s' --mode=json --warning-rate-limit=%i
--limit-msg-len=%i > '%s'"

class PluginProps:
...

Do we need *two empty lines* between FILTER_CMD and class PluginsProps?

Thanks, Leo

lzaoral:
Thanks for the question, @lbossis!  Yes, we need two empty lines to stay conformant to the office Python Style Guide ([PEP8](https://peps.python.org/pep-0008/)).  The section about blank lines explicitly says:

> [Blank Lines](https://peps.python.org/pep-0008/#blank-lines)
  Surround top-level function and class definitions with two blank lines.
  ...

kdudka:
@lbossis Thanks for the update!  Could you please update the commit message summary as requested at https://github.com/csutils/csmock/pull/114#issuecomment-1653170888 ?

The `gitleaks: ` prefix is crucial here because the changes are specific to a single plug-in of csmock.

kdudka:
@lbossis Thanks for the update!  Unfortunately the commit message needs to be reformatted like this:
```
gitleaks: implement --gitleaks-limit-msg-len defaulting to 512

This package contains an extension of gitleaks plug-in to support a new
command line option for controlling event message length from csmock.
csmock-plugin-gitleaks is dependent on the updated version of csdiff

Fixes: https://github.com/csutils/csdiff/issues/114
Related: https://issues.redhat.com/browse/OSH-67
```

The git commit summary is the first line of the commit message, delimited by a blank line from the description.  The commit message summary has a special semantics in git.  See the `git-commit(1)` man page:
```
Though not required, it’s a good idea to begin the commit message with a single short (less than 50 character) line summarizing the change, followed by a blank line and then a more thorough description. The text up to the first blank line in a commit message is treated as the commit title, and that title is used throughout Git. For example, git-format-patch(1) turns a commit into email, and it uses the title on the Subject line and the rest of the commit in the body.
```
You can easily update the commit message with: `git commit --amend && git push --force`

lbossis:
Commit message has been updated according to the recommended format.

lzaoral (approved):

kdudka:
The original commit message referenced an unrelated csdiff issue by mistake (https://github.com/csutils/csdiff/issues/114) so I have fixed it to make it reference this pull request instead.  Merging...

PR #116, opened 2023-08-24
Fix missing dependency

rhyw:
The `bld2repo` executive is being used, but corresponding package
`modulemd-tools` was not declared as a dependency earlier.

### DIFF
--- a/make-srpm.sh
+++ b/make-srpm.sh
@@ -128,6 +128,9 @@ Requires: csdiff > 3.0.3
 Requires: csgcca
 Requires: cswrap
 Requires: mock
+%if 0%{?rhel} != 7
+Recommends: modulemd-tools
+%endif
 
 %description -n csmock-common
 This package contains the csmock tool that allows to scan SRPMs by Static


kdudka (changes requested):
I can see that EPEL-7 builds in CI started to fail with:
```
error: line 53: Unknown tag: Recommends: modulemd-tools
```

... and there is no point in recommending `modulemd-tools` on RHEL-7/CentOS-7 anyway because there is no officially supported `modulemd-tools` package on such systems.  Could you please make the recommendation conditional like this?
```diff
--- a/make-srpm.sh
+++ b/make-srpm.sh
@@ -125,13 +125,15 @@ Tool for plugging static analyzers into the build process, free of mock.
 %package -n csmock-common
 Summary: Core of csmock (a mock wrapper for Static Analysis tools)
 Requires: csdiff > 3.0.3
 Requires: csgcca
 Requires: cswrap
 Requires: mock
+%if 0%{?rhel} != 7
 Recommends: modulemd-tools
+%endif

 %description -n csmock-common
 This package contains the csmock tool that allows to scan SRPMs by Static
 Analysis tools in a fully automated way.

 %package -n csmock-plugin-bandit
```

rhyw:
Updated as per suggestions.

kdudka (approved):
Merging, thanks!

PR #118, opened 2023-08-29
csmock: drop `DEFAULT_KNOWN_FALSE_POSITIVES_FALLBACK`

kdudka:
The default path was renamed more than 2 years ago.  Nobody should use the fallback path any more.

Related: commit d7afe41f351c9dc18d71fe21683342bedba7d7ce


lzaoral (approved):
LGTM!

PR #124, opened 2023-09-26
common: present important results by default

kdudka:
... if any filters of important results are defined.

The full results are now available with the `-all` suffix.  The files with `-imp` suffix are provided as symlinks to the filtered results for compatibility.

Resolves: https://issues.redhat.com/browse/OSH-299

### DIFF
--- a/py/common/results.py
+++ b/py/common/results.py
@@ -327,7 +327,8 @@ def finalize_results(js_file, results, props):
         tmp_js_file = re.sub("\\.js", "-tmp.js", js_file)
         cmd = "cslinker --implist '%s' '%s' > '%s' && mv -v '%s' '%s'" \
                 % (imp_js_file, js_file, tmp_js_file, tmp_js_file, js_file)
-        results.exec_cmd(cmd, shell=True)
+        if 0 != results.exec_cmd(cmd, shell=True):
+            results.error("failed to tag important findings in the full results", ec=0)
 
     (err_file, _) = transform_results(js_file, results)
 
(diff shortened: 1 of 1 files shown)


py/common/results.py
```diff
@@ -320,15 +320,22 @@ def finalize_results(js_file, results, props):
...
 
-        # generate *-imp.{err,html}
-        transform_results(imp_js_file, results)
-
-        # initialize the "imp" flag in the resulting full .js output file
-        tmp_js_file = re.sub("\\.js", "-tmp.js", js_file)
+        # initialize the "imp" flag in the resulting `-all.js` output file
+        # and replace the original .js file by `-imp.js`
+        all_js_file = re.sub("\\.js", "-all.js", js_file)
         cmd = "cslinker --implist '%s' '%s' > '%s' && mv -v '%s' '%s'" \
```
jperezdealgaba, line 326: Although this line wasn't "directly" modified, shouldn't we check that the process finished correctly?

kdudka: Good question.  The exit code of `cslinker` is in fact checked by the `&&` shell construction -- if it fails, the original file is not overwritten by `mv`.  Similarly with `mv`, which works atomically -- if it fails, the original file is not overwritten.  Nevertheless, we can emit a non-voting error to make it more obvious in `scan.log` if it happens.

kdudka: Fixed in commit 2637404448a40682e2b7adedb3786100d8f41c7f.

jperezdealgaba (approved):
/lgtm

jperezdealgaba:
I have approved the changes

kdudka:
@jperezdealgaba Thanks for review!

PR #127, opened 2023-10-06
snyk: wrap snyk invocation by `timeout(1)` by default

kdudka:
... set to 5 hours so that Snyk does not hang forever and does not block OpenScanHub scanning queue.  If Snyk does not deliver results within the first 5 hours, it usually hangs forever.

The amount of time can be tweaked by the `--snyk-timeout` option of csmock or disabled by `--snyk-timeout=0`.

Related: https://issues.redhat.com/browse/OSH-301

### DIFF
--- a/py/plugins/snyk.py
+++ b/py/plugins/snyk.py
@@ -32,6 +32,9 @@ SNYK_LOG = "/builddir/snyk-capture.log"
 
 FILTER_CMD = f"csgrep '%s' --mode=json --prepend-path-prefix={SNYK_SCAN_DIR}/ > '%s'"
 
+# default value for the maximum amount of time taken by invocation of Snyk (5 hours)
+DEFAULT_SNYK_TIMEOUT=18000
+
 
 class PluginProps:
     def __init__(self):
@@ -66,6 +69,11 @@ class Plugin:
         parser.add_argument(
             "--snyk-refresh", action="store_true",
             help="force download of snyk binary executable")
+
+        parser.add_argument(
+            "--snyk-timeout", type=int, default=DEFAULT_SNYK_TIMEOUT,
+            help="maximum amount of time taken by invocation of Snyk [s]")
+
 
     def handle_args(self, parser, args, props):
         if not self.enabled:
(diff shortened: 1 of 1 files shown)


py/plugins/snyk.py
```diff
@@ -32,6 +32,9 @@
 
 FILTER_CMD = f"csgrep '%s' --mode=json --prepend-path-prefix={SNYK_SCAN_DIR}/ > '%s'"
 
+# default value for the maximum amount of time taken by invocation of Snyk (5 hours)
+DEFAULT_SNYK_TIMEOUT=18000
```
jperezdealgaba, line 36: @kdudka Hey Kamil. I don't know if because of the workers, the scans are slower. But when I have been testing this, more than 2 hours means it is going to get stuck. Just in case you want to reduce this value.

jperezdealgaba (approved):
/lgtm

py/plugins/snyk.py
```diff
@@ -147,9 +155,15 @@ def scan_hook(results, mock, props):
...
                 return ec
 
-            # run snyk code
+            # command to run snyk code
             cmd = "%s code test -d %s --sarif-file-output=%s >/dev/null 2>%s" \
                     % (self.snyk_bin, SNYK_SCAN_DIR, SNYK_OUTPUT, SNYK_LOG)
+
+            if args.snyk_timeout:
+                # wrap snyk invocation by timeout(1)
+                cmd = f"/usr/bin/timeout {args.snyk_timeout} {cmd}"
```
jperezdealgaba, line 164: @kdudka Shouldn't you have an "else" clause to run the command if snyk_timeout it is not set?

jperezdealgaba:
@kdudka I approved the changes but then noticed the "else" clause. Please, check it before merging it

py/plugins/snyk.py
kdudka: Yes, that was my observation as well when I tested Snyk initially (although the scans were never hanging when I was experimenting with the Snyk client interactively).  But now we are increasing the load on the 3rd party scanning service.  Theoretically up to 40 Snyk scans can now run in parallel in OpenScanHub.  So I wanted to reserve some extra time to avoid failing OSH tasks just because a scan is temporarily queued on the remote service.

The 5 hours timeout is really a workaround to avoid manual intervention in the edge case where the Snyk client keeps hanging forever.  This should not happen and the vendor should fix the client to terminate itself in such cases because the client has more information to decide whether it makes sense to wait longer or not.  For example, if the remote scanning process crashes, it would be wasteful to wait even for 2 hours before we terminate the client.  However, if the scan is just temporarily queued on the remote service, then waiting for 3 hours might be better for us than failing the OSH task and having to manually investigate what happened.

jperezdealgaba: Cool then. I have used the same timout for cspodman

PR #128, opened 2023-10-07
csmock: implement `--warning-rate-limit` and `--limit-msg-len`

kdudka:


py/csmock
```diff
@@ -1227,15 +1246,32 @@
...
+                        csgrep_cmd += " " + strlist_to_shell_cmd(props.results_limits_opts)
+
+                    cmd = strlist_to_shell_cmd(mock.get_mock_cmd(["--shell", csgrep_cmd]))
+                    cmd += f" <'{all_file}' >'{tmp_file}'"
                     if results.exec_cmd(cmd, shell=True) == 0:
                         shutil.move(tmp_file, all_file)
+                        props.results_limits_applied = True
 
             # we are done with mock
+
```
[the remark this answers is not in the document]
kdudka: Fixed in commit 0be37cd14aeb579e343f68757888511d369ae0f8.

py/csmock
```diff
@@ -1227,16 +1246,33 @@ cd %%s*/ || cd *\n\
...
 
+            # make sure to apply results limits because `csgrep --embed-context` might not be available in chroot
+            if props.results_limits_opts and not props.results_limits_applied:
+                tmp_file = f"{all_file}.tmp"
+                csgrep_cmd = "csgrep --mode=json " + strlist_to_shell_cmd(props.results_limits_opts)
+                csgrep_cmd += f" '{all_file}' >'{tmp_file}'"
+                ec = results.exec_cmd(csgrep_cmd, shell=True)
+                if 0 == ec:
+                    shutil.move(tmp_file, all_file)
+                    props.results_limits_applied = True
```
hanchuntao, line 1272: I don't understand why `if props.results_limits_opts and not props.results_limits_applied`, and `if 0 == ec`, then need to set `props.results_limits_applied = True`. Could you provide more detail?

kdudka: The value set on this line is currently not used by csmock.  It is just to make `props.results_limits_applied` reflect the current state (because we know that results limits have been successfully applied).  The `props` object is passed to all plug-in hooks.  In theory, a csmock plug-in (or even csmock itself) may use the value for something in the future, so I wanted to make sure that the value is accurate.

hanchuntao: I got it, thanks!

hanchuntao:
LGTM

hanchuntao (approved):

kdudka:

PR #129, opened 2023-10-09
packaging: add `tar` and `xz` to `Requires`

lzaoral:
Fixes the following error on minimal OSH worker installations:
```
<<< 2023-10-09 13:27:08	csmock exit code: 0

/bin/sh: line 1: xz: command not found
tar: /tmp/tmpa05fux7j/units-2.21-5.fc37.tar.xz: Wrote only 4096 of 10240 bytes
tar: Child returned status 127
tar: Error is not recoverable: exiting now
!!! 2023-10-09 13:27:08	fatal error: failed to write '/tmp/tmpa05fux7j/units-2.21-5.fc37.tar.xz', not removing '/tmp/csmockbyw9qj7l'...
```

### DIFF
--- a/make-srpm.sh
+++ b/make-srpm.sh
@@ -128,6 +128,8 @@ Requires: csdiff > 3.0.4
 Requires: csgcca
 Requires: cswrap
 Requires: mock
+Requires: tar
+Requires: xz
 %if 0%{?rhel} != 7
 Recommends: modulemd-tools
 %endif


make-srpm.sh
```diff
@@ -105,6 +105,9 @@ Requires: csmock-plugin-cppcheck        >= %{version}-%{release}
 Requires: csmock-plugin-gitleaks        >= %{version}-%{release}
 Requires: csmock-plugin-shellcheck      >= %{version}-%{release}
 
+Requires: tar
+Requires: xz
```
kdudka, line 109: These should go to the `csmock-common` sub-package.

kdudka (changes requested):

make-srpm.sh
lzaoral: Addressed in 00541f1f5241f44e99483e1455e8f4fc9e54eb17.

kdudka (approved):
Looks good.  Thanks!

PR #131, opened 2023-10-10
csmock-snyk-plugin: no supported projects fix

jperezdealgaba:
Fix: https://issues.redhat.com/browse/OSH-329

Reproducer: csmock -r fedora-rawhide-x86_64 -t snyk -o /tmp/tmp_yvvqrnh/attache.tar.xz  --gcc-analyze --unicontrol-notests --unicontrol-bidi-only --shell-cmd=:'/tmp/tmp_yvvqrnh/attache.tar.gz'

Attache needs to be downloaded from: https://github.com/RedHatInsights/attache

If no supported project is detected using the snyk plugin, it no longer crashes but it does return 0.

### DIFF
--- a/py/plugins/snyk.py
+++ b/py/plugins/snyk.py
@@ -170,6 +170,10 @@ class Plugin:
             mock.exec_chroot_cmd("/bin/rm -fv %s" % auth_token_dst)
 
             # check exit code of snyk code itself
+            if ec == 3:
+                # If there are no supported project, we return no results but no crash.
+                results.print_with_ts("snyk-code: no supported projects detected")
+                return 0
             if ec not in [0, 1]:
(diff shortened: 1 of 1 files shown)


py/plugins/snyk.py
```diff
@@ -170,6 +170,10 @@ def scan_hook(results, mock, props):
             mock.exec_chroot_cmd("/bin/rm -fv %s" % auth_token_dst)
 
             # check exit code of snyk code itself
+            if ec == 3:
+                 # If there are no supported project, we return no results but no crash.
```
kdudka, line 174: Please remove the extra space before `#`.

kdudka (approved):

py/plugins/snyk.py
jperezdealgaba: Just did it

PR #132, opened 2023-10-17
csmock-snyk-plugin: no supported projects fix

jperezdealgaba:
Fix: https://issues.redhat.com/browse/OSH-329

Reproducer: csmock -t snyk -r rhel-8-x86_64 --shell-cmd=: <attache.tar.gz tarball>

Attache needs to be downloaded from https://github.com/RedHatInsights/attache

If no supported project is detected using the snyk plugin, it no longer crashes and no results are parsed

### DIFF
--- a/py/plugins/snyk.py
+++ b/py/plugins/snyk.py
@@ -172,7 +172,9 @@ class Plugin:
             # check exit code of snyk code itself
             if ec == 3:
                 # If there are no supported project, we return no results but no crash.
-                results.print_with_ts("snyk-code: no supported projects detected")
+                results.print_with_ts("snyk-code: no supported project for Snyk")
+                # If no supported project, no results file is generated
+                props.copy_out_files.remove(SNYK_OUTPUT)
                 return 0
             if ec not in [0, 1]:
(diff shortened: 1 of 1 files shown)


py/plugins/snyk.py
```diff
@@ -172,7 +172,10 @@ def scan_hook(results, mock, props):
             # check exit code of snyk code itself
             if ec == 3:
                 # If there are no supported project, we return no results but no crash.
-                results.print_with_ts("snyk-code: no supported projects detected")
+                results.print_with_ts("snyk-code: no supported project for Snyk")
+                # If no supported project, no results file is generated
+                if SNYK_OUTPUT in props.copy_out_files:
```
kdudka, line 177: The condition is redundant.  You can just call `props.copy_out_files.remove(SNYK_OUTPUT)` unconditionally with the exactly same effect.

kdudka (changes requested):
Please also remove the unnecessary merge commit that sneaked in.

kdudka (approved):
Looks good.  Thanks!

PR #133, opened 2023-10-19
snyk: do not throw away Snyk results by mistake

kdudka:
... on successful scans.  This commit fixes a major regression introduced by the previous commit.

Fixes: commit e8863429bd61fd5308547b084c14f23ac8432c58
Related: https://issues.redhat.com/browse/OSH-329
Resolves: https://issues.redhat.com/browse/OSH-362

### DIFF
--- a/py/plugins/snyk.py
+++ b/py/plugins/snyk.py
@@ -187,10 +187,10 @@ class Plugin:
 
         # convert the results into the csdiff's JSON format
         def filter_hook(results):
-            # If no supported project is found, we don't have results file
-            if not os.path.exists(SNYK_OUTPUT):
-                return 0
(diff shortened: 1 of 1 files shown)


jperezdealgaba (approved):
/lgtm

lzaoral:
@kdudka @jperezdealgaba Thank you both for the quick fix!

It may be a good idea to extend the changes done in #95 with a follow up PR to also test the Snyk `csmock` plug-in so that we can prevent such breakages in the future.

Also, @jperezdealgaba, please, make **sure** to test your contributions properly.   I'm sure that you have tested that the original PR fixed the crash reproducer but it is a **very** good practice to also check that the basic common use cases won't be affected by your changes and that they still work and produce reasonable results.  Thank you!

PR #134, opened 2023-10-19
snyk: unify the names of intermediate files

kdudka:
... to match the file names used by cspodman.  It is utterly confusing for users and developers of csmock and cspodman if each tool uses different file names for the same things.

This divergence somehow slipped through the reviews on cspodman, which implemented the support for Snyk later.  This commit makes it consistent again.

Closes: https://github.com/csutils/csmock/pull/134


jperezdealgaba (approved):
/lgtm

PR #137, opened 2023-11-08
snyk-plugin: added --snyk-code-test-opts flag

jperezdealgaba:
Users are now able to use the introduced flag to pass new parameters to the snyk cli

Resolves: https://issues.redhat.com/browse/OSH-307

csmock -t snyk --snyk-code-test-opts="--report --project-name=osbuild" -r rhel-8-x86_64 osbuild-99-1.el10+1.src.rpm

Screenshot 2023-11-08 at 11 07 28

### DIFF
--- a/py/plugins/snyk.py
+++ b/py/plugins/snyk.py
@@ -33,7 +33,7 @@ SNYK_LOG = "/builddir/snyk-scan.log"
 FILTER_CMD = f"csgrep '%s' --mode=json --prepend-path-prefix={SNYK_SCAN_DIR}/ > '%s'"
 
 # default value for the maximum amount of time taken by invocation of Snyk (5 hours)
-DEFAULT_SNYK_TIMEOUT=18000
+DEFAULT_SNYK_TIMEOUT = 18000
 
 
 class PluginProps:
@@ -74,6 +74,9 @@ class Plugin:
             "--snyk-timeout", type=int, default=DEFAULT_SNYK_TIMEOUT,
             help="maximum amount of time taken by invocation of Snyk [s]")
 
+        parser.add_argument(
+            "--snyk-code-test-opts",
+            help="space-separated list of additional options passed to the 'snyk code test' command")
 
     def handle_args(self, parser, args, props):
         if not self.enabled:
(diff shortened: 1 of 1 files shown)


py/plugins/snyk.py
```diff
@@ -177,7 +186,7 @@ def scan_hook(results, mock, props):
                 props.copy_out_files.remove(SNYK_OUTPUT)
                 return 0
             if ec not in [0, 1]:
-                results.error("snyk code returned unexpected exit status: %d" % ec, ec=ec)
+                results.error(f"snyk code returned unexpected exit status: {ec}", ec=ec)
```
jperezdealgaba, line 189: This line just modified in order to pass the tests. As they were failing because of an f-string

py/plugins/snyk.py
```diff
@@ -17,7 +17,6 @@
 
 import os
 
-
```
kdudka, line 20: This line should not be removed.

py/plugins/snyk.py
```diff
@@ -103,7 +105,8 @@ def fetch_snyk_hook(results, props):
             self.snyk_bin = os.path.join(cache_dir, snyk_bin_name)
 
             if not args.snyk_refresh and os.path.exists(self.snyk_bin):
-                results.print_with_ts("reusing previously downloaded snyk executable: %s" % self.snyk_bin)
+                results.print_with_ts("reusing previously downloaded snyk executable: "
+                                      "%s" % self.snyk_bin)
```
kdudka, line 109: This does not improve code readability at all.  Please use ` + self.snyk_bin` like we do in `cspodman`.

py/plugins/snyk.py
```diff
@@ -103,7 +105,8 @@ def fetch_snyk_hook(results, props):
             self.snyk_bin = os.path.join(cache_dir, snyk_bin_name)
 
             if not args.snyk_refresh and os.path.exists(self.snyk_bin):
-                results.print_with_ts("reusing previously downloaded snyk executable: %s" % self.snyk_bin)
+                results.print_with_ts("reusing previously downloaded snyk "
+                                      "executable: %s" % self.snyk_bin)
```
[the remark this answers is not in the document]
kdudka: I am not sure what happened in this specific case but matching previously reported findings in updated code is an open problem.  The tool cannot check line numbers for equality because those are moved up/down by the code changes.  Consequently, if two findings are indistinguishable from each other, a valid finding may be reported with a wrong line number in the list of newly introduced findings: https://github.com/csutils/csdiff/blob/f482df6295f27eeb2ca346fab5b6a233f97d7e55/src/lib/deflookup.cc#L115

py/plugins/snyk.py
kdudka: @jperezdealgaba Could you please rebase this onto `main` to eliminate the Pylint complaints about long lines?

jperezdealgaba: Sure, no problem! Just did it!

kdudka:
@jperezdealgaba Thanks for the update!  Merging...

PR #138, opened 2023-11-09
results: reset imp flag in the list of important findings only

kdudka:
Tagging important findings does not make sense in the list of important findings only.

Resolves: https://issues.redhat.com/browse/OSH-343
Depends-on: https://github.com/csutils/csdiff/pull/149


hanchuntao (approved):
LGTM

kdudka:

PR #148, opened 2024-01-05
results: print `warning` for recoverable errors

kdudka:
... to make it obvious in `scan.log` that they did not affect the overall exit code.

### DIFF
--- a/py/common/results.py
+++ b/py/common/results.py
@@ -181,7 +181,8 @@ class ScanResults:
             self.ec = ec
 
     def error(self, msg, ec=1, err_prefix=""):
-        self.print_with_ts("%serror: %s\n" % (err_prefix, msg), prefix="!!! ")
+        level = "warning" if ec == 0 else "error"
+        self.print_with_ts(f"{err_prefix}{level}: {msg}\n", prefix="!!! ")
         self.update_ec(ec)
         if not self.dying and not self.keep_going and (self.ec != 0):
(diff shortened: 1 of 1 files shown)


py/common/results.py
```diff
@@ -181,7 +181,8 @@ def update_ec(self, ec):
             self.ec = ec
 
     def error(self, msg, ec=1, err_prefix=""):
-        self.print_with_ts("%serror: %s\n" % (err_prefix, msg), prefix="!!! ")
+        level = "warning" if ec == 0 else "error"
```
rhyw, line 184: I'm a bit confused to the conventions, e.g., to have `results.error` with `level=warning`...

Anyway it looks good to me, since there are cases in existing code where we use `results.error` together with `ec=0`

kdudka: @rhyw The interface is inspired by the [error(3)](https://man7.org/linux/man-pages/man3/error.3.html) function from the standard C library:

> If  status  has a nonzero value, then error() calls exit(3) to terminate the program using the given value as the exit status; otherwise it returns after printing the error message.

rhyw:
LGTM

kdudka:
@rhyw Please approve it in the GitHub UI so that I can merge it.

rhyw (approved):

rhyw:
> @rhyw Please approve it in the GitHub UI so that I can merge it.

Done.

PR #149, opened 2024-01-08
plugins/semgrep: add csmock semgrep plugin

rhyw:
Resolves: https://issues.redhat.com/browse/OSH-57
Closes: https://github.com/csutils/csmock/pull/149


py/plugins/semgrep.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def handle_args(self, parser, args, props):
+        if not self.enabled:
+            return
+
+        # download semgrep rules from downstream repo
+        def fetch_semgrep_rules_hook(results, props):
+            # use results.exec_cmd to install semgrep cli using pip
+            cmd = "python3 -m pip install --user semgrep"
+            ec = results.exec_cmd(cmd, shell=True)
```
kdudka, line 84: A csmock plug-in should not silently install anything in the host environment.  If it needs to download a tool from the Internet, it should use a separate cache directory that will not affect user's environment.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,166 @@
...
+                # remove previously downloaded semgrep repo/rules
+                results.exec_cmd(["rm", "-rf", f"{SEMGREP_RULES_CACHE_DIR}/*"])
+
+                # fetch semgrep rules from downstream repo
+                ec = results.exec_cmd(repo_clone_cmd.split())
+                if 0 != ec:
+                    results.error("failed to fetch semgrep rules from downstream repo")
+                    return ec
+            # query version of semgrep
+            semgrep_bin_dir = os.path.abspath("./.local/bin")
```
kdudka, line 110: Does the code expect `csmock` to be called in a specific directory?

py/plugins/semgrep.py
```diff
@@ -0,0 +1,166 @@
...
+            # run semgrep scan
+            ec = mock.exec_chroot_cmd(cmd)
+
+            # below are all possible return codes from semgrep scan
+            if ec == 123:
+                results.error("Indiscriminate errors reported on standard error.")
+            elif ec == 124:
+                results.error("Command line parsing errors.")
+            elif ec == 125:
+                results.error("Unexpected internal errors (bugs).")
```
kdudka, line 151: These messages should ideally use the `semgrep: ` prefix to make it clear what failed from the error message itself.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,166 @@
...
+        # run semgrep scan after successful build
+        props.post_install_hooks += [scan_hook]
+
+        # convert the results into the csdiff's JSON format
+        def filter_hook(results):
+            src = results.dbgdir_raw + SEMGREP_SCAN_OUTPUT
+            if not os.path.exists(src):
+                return 0
+            dst = "%s/semgrep-scan-results.json" % results.dbgdir_uni
+            cmd = FILTER_CMD % (src, dst)
```
kdudka, line 163: Please use f-strings in newly introduced code.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,166 @@
...
+            if 0 != ec:
+                results.error("failed to query semgrep cli version", ec=ec)
+                return ec
+
+            # parse and record the version of semgrep cli
+            version = out.split(" ")[0]
+            results.ini_writer.append("analyzer-version-semgrep-cli", version)
+
+            # copy semgrep rules into the chroot
+            props.copy_in_files += [SEMGREP_RULES_CACHE_DIR]
```
kdudka, line 123: This will copy the rules into chroot but what about the semgrep tool itself?

py/plugins/semgrep.py
```diff
@@ -0,0 +1,166 @@
...
+                   f" --config={SEMGREP_RULES_CACHE_DIR}/rules")
+
+            # append additional options passed to the 'semgrep scan' command
+            if args.semgrep_scan_opts:
+                cmd += f" {args.semgrep_scan_opts}"
+
+            # eventually append the target directory to be scanned
+            cmd += f" --output={SEMGREP_SCAN_OUTPUT} {SEMGREP_SCAN_DIR} 2>{SEMGREP_SCAN_LOG}"
+            # run semgrep scan
+            ec = mock.exec_chroot_cmd(cmd)
```
kdudka, line 143: This will run `semgrep` in the chroot environment after the tool was installed into the host environment.  How could this ever work?

kdudka (changes requested):

py/plugins/semgrep.py
rhyw: Updated

py/plugins/semgrep.py
rhyw: Latest code updated to specify the install target.

py/plugins/semgrep.py
rhyw: This was updated, by updating one-time `PATH`/`PYTHONPATH`.

py/plugins/semgrep.py
rhyw: Fixed

py/plugins/semgrep.py
rhyw: rules should be located in host, this line was removed

py/plugins/semgrep.py
rhyw: Updated

py/plugins/semgrep.py
```diff
@@ -0,0 +1,189 @@
...
+            try:
+                # make sure the cache directory exists
+                os.makedirs(SEMGREP_RULES_CACHE_DIR, mode=0o755, exist_ok=True)
+                os.makedirs(SEMGREP_LIB_TARGET_DIR, mode=0o755, exist_ok=True)
+            except OSError:
+                results.error("failed to create semgrep rules/lib directory")
+                return 1
+
+            # install semgrep cli using pip
+            cmd = f"python3 -m pip install --target={SEMGREP_LIB_TARGET_DIR} semgrep"
```
kdudka, line 103: What will happen if we start 5 processes of `csmock -t semgrep ...` at the same time?  Will the shared cache directory always be in a consistent state after executing this command?

py/plugins/semgrep.py
```diff
@@ -0,0 +1,189 @@
...
+
+            # command to fetch semgrep rules
+            repo_clone_cmd = f"git clone {args.semgrep_rules_repo} {SEMGREP_RULES_CACHE_DIR}"
+
+            # check whether we can reuse previously downloaded semgrep rules
+            if not args.semgrep_rules_refresh and os.listdir(SEMGREP_RULES_CACHE_DIR):
+                results.print_with_ts(f"reusing previously downloaded semgrep rules: {SEMGREP_RULES_CACHE_DIR}")
+            else:
+                # remove previously downloaded semgrep repo/rules
+                results.exec_cmd(["rm", "-rf", f"{SEMGREP_RULES_CACHE_DIR}/*"])
```
kdudka, line 119: This will not work well if we run multiple `csmock -t semgrep ...` processes in parallel.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,189 @@
...
+
+import os
+
+# disable metrics to be sent to semgrep cloud
+SEMGREP_SEND_METRICS = "off"
+
+SEMGREP_VERBOSE = True
+
+# FIXME: should we use a dedicated repo to maintain semgrep rules?
+SEMGREP_RULES_REPO = "https://git.prodsec.redhat.com/prodsec-services/masssast"
```
kdudka, line 26: This URL is not publicly reachable and it is definitely not a good default for the upstream csmock project.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,189 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with csmock.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+
+# disable metrics to be sent to semgrep cloud
+SEMGREP_SEND_METRICS = "off"
```
kdudka, line 21: The variables setting default values for options should start with the `DEFAULT_` prefix.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,189 @@
...
+
+            # check whether we can reuse previously downloaded semgrep rules
+            if not args.semgrep_rules_refresh and os.listdir(SEMGREP_RULES_CACHE_DIR):
+                results.print_with_ts(f"reusing previously downloaded semgrep rules: {SEMGREP_RULES_CACHE_DIR}")
+            else:
+                # remove previously downloaded semgrep repo/rules
+                results.exec_cmd(["rm", "-rf", f"{SEMGREP_RULES_CACHE_DIR}/*"])
+
+                # fetch semgrep rules from downstream repo
+                ec = results.exec_cmd(repo_clone_cmd.split())
```
kdudka, line 122: Why do you use a format string to create a space-separated list to convert it to a list later on?  This would unnecessarily break e.g. if `SEMGREP_RULES_CACHE_DIR` contained a space.  It is much easier to read (and more robust at the same time) if you create the list directly:
```py
repo_clone_cmd = ["git", "clone", args.semgrep_rules_repo, SEMGREP_RULES_CACHE_DIR]
ec = results.exec_cmd(repo_clone_cmd)
```

py/plugins/semgrep.py
```diff
@@ -0,0 +1,189 @@
...
+
+            # get the results out of the chroot
+            props.copy_out_files += [SEMGREP_SCAN_OUTPUT, SEMGREP_SCAN_LOG]
+            return 0
+
+        props.pre_mock_hooks += [fetch_semgrep_rules_hook]
+
+        def get_chroot_root_path(results, props):
+            # get the path to the chroot root directory
+            cmd = f"mock -r {props.mock_profile} --print-root-path"
```
kdudka, line 144: This needs to go through the `MockWrapper` abstraction.  Otherwise it would not work with `csmock --root-override ...` etc.:
```py
cmd = mock.get_mock_cmd(["--print-root-path"])
(ec, out) = results.get_cmd_output(cmd, shell=False)
```

py/plugins/semgrep.py
```diff
@@ -0,0 +1,189 @@
...
+            version = output.rstrip("\n")
+            results.ini_writer.append("analyzer-version-semgrep-cli", version)
+
+            # get the results out of the chroot
+            props.copy_out_files += [SEMGREP_SCAN_OUTPUT, SEMGREP_SCAN_LOG]
+            return 0
+
+        props.pre_mock_hooks += [fetch_semgrep_rules_hook]
+
+        def get_chroot_root_path(results, props):
```
kdudka, line 142: Why is this a local function when it does not access anything local to `handle_args()`?

kdudka (changes requested):

py/plugins/semgrep.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def handle_args(self, parser, args, props):
+        if not self.enabled:
+            return
+
+        def populate_semgrep_runtime_env():
+            # Update PYTHONPATH and PATH, so that semgrep cli and its dependent libraries can be found
+            # Note that this change doesn't persist across sessions.
+            os.environ['PYTHONPATH'] = f"{os.environ.get('PYTHONPATH')}:{SEMGREP_LIB_TARGET_DIR}"
+            os.environ['PATH'] = f"{os.environ.get('PATH')}:{SEMGREP_LIB_TARGET_DIR}/bin"
```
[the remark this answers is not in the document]
kdudka: As pointed out by @lzaoral, this will also prepend `None` to `${PYTHONPATH}` if the variable is initially unset.

make-srpm.sh
```diff
@@ -343,6 +350,10 @@ This package contains the unicontrol plug-in for csmock.
 %{python3_sitelib}/csmock/plugins/snyk.py*
 %{python3_sitelib}/csmock/plugins/__pycache__/snyk.*
 
+%files -n csmock-plugin-semgrep
```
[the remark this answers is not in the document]
rhyw: updated

py/CMakeLists.txt
```diff
@@ -66,6 +66,7 @@ if(ENABLE_CSMOCK)
     install_plugin(shellcheck)
     install_plugin(smatch)
     install_plugin(snyk)
+    install_plugin(semgrep)
```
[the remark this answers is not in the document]
rhyw: updated

py/plugins/semgrep.py
rhyw: updated

py/plugins/semgrep.py
```diff
@@ -0,0 +1,189 @@
...
+
+class PluginProps:
+    def __init__(self):
+        self.description = (
+            "A fast, open-source, static analysis engine for finding bugs, "
+            "detecting dependency vulnerabilities, and enforcing code standards."
+        )
+
+        # include this plug-in in `csmock --all-tools`
+        self.stable = True
```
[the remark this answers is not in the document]
rhyw: updated

py/plugins/semgrep.py
rhyw: What do you suggest? by add a new repo in github under `csutils` and maintain those rules on our own?

py/plugins/semgrep.py
rhyw: I don't think so. it's indeed unsafe if it's ran in parallel. I don't have a better idea how this could be done, unless the semgrep executable is installed before scan begins. i.e., `semgrep` as a install require while deploying the worker instead of installing at runtime. Is that an option?

py/plugins/semgrep.py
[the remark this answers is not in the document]
rhyw: This only applies to current process. Other process won't be impacted.

py/plugins/semgrep.py
kdudka: If there is any well-maintained publicly available repository, we can use it as the default value.  If not, we should not provide any default URL and make this option mandatory so that the user provides it (assuming `semgrep` does not work without it).

py/plugins/semgrep.py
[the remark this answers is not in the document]
kdudka: I do not understand what you are trying to say.  What do you mean by `current process`?  `csmock`,  `semgrep`, or something else?  It does not make any sense...

py/plugins/semgrep.py
kdudka: Let's start with creating something that works reliably.  Optimizations can be introduced later as needed.  If we install `semgrep` into `results.tmpdir` without caching anything across `csmock` runs, it will be safe regarding parallel execution of `csmock`.

rhyw: sure. I'll update to make sure it's installed under `results.tmpdir`.

py/plugins/semgrep.py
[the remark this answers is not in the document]
rhyw: The problem is that after the package is installed in a user-specified directory, the `bin`/`lib` directory is invisible, thus `PATH`/`PYTHONPATH` has to be amended to make it visible. 

> The environment should be changed only for commands invoked by this plug-in, rather than for all commands invoked by `csmock`.

I'm not sure I understand what this means. How am I supposed to make it visible without changing like this? Sorry for the confusion, I don't have a good understanding of how it works to be frank.

py/plugins/semgrep.py
rhyw: The offical semgrep-rules repo is well maintained, but it's unfortunate it cannot be used directly, i.e., I'll need to specify rules one by one, instead of by ruleset.

I tried several other open source rules repo, none seem to be qualified. I'll see if I could find one, otherwise I'll make it mandatory option. Didn't know that before till I see your comment.

py/plugins/semgrep.py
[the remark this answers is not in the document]
kdudka: You only need to set variables when you invoke `semgrep`.

If you use shell (like in the `semgrep scan ...` command), you can just prepend the variables assignment to the command you want to execute:
```py
cmd = f"PYTHONPATH=\"{semgrep_lib_dir}:$PYTHONPATH\" {cmd}"
cmd = f"PATH=\"{semgrep_lib_dir}/bin:$PATH\" {cmd}"
```

If you use `exec` directly (like in the `["semgrep", "--version"]` command), you can wrap the command by `env(1)` to set the environment:
```py
cmd = ["semgrep", "--version"]
cmd = ["env", f"PATH={path_for_semgrep}", f"PYTHONPATH={pythonpath_for_semgrep}"] + cmd
```

I would suggest to use one approach for both the commands so that the code for env handling can be shared.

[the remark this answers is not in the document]
rhyw: Thanks. I'll update this way!

py/plugins/semgrep.py
rhyw: I use it as a shortcut to get chroot root path. anyway it's not really necessary. Removed.

py/plugins/semgrep.py
rhyw: updated

py/plugins/semgrep.py
```diff
@@ -0,0 +1,168 @@
...
+    def enable(self):
+        self.enabled = True
+
+    def init_parser(self, parser):
+        parser.add_argument(
+            "--semgrep-metrics", default=DEFAULT_SEMGREP_SEND_METRICS,
+            help="configures how usage metrics are sent to the Semgrep server")
+
+        parser.add_argument(
+            "--semgrep-rules-repo", required=True,
```
kdudka, line 61: You cannot require the option this way because it would fail all scans where `semgrep` is not enabled (as long as this csmock plug-in is installed).

py/plugins/semgrep.py
```diff
@@ -0,0 +1,168 @@
...
+
+            # install semgrep cli using pip
+            cmd = f"python3 -m pip install --target={semgrep_lib_dir} semgrep=={SEMGREP_CLI_VERSION}"
+            ec = results.exec_cmd(cmd, shell=True)
+            if 0 != ec:
+                results.error("failed to install semgrep cli using pip")
+
+            semgrep_prefix = f"env PATH={semgrep_lib_dir}/bin:$PATH, PYTHONPATH={semgrep_lib_dir}"
+
+            repo_clone_cmd = ["git", "clone", args.semgrep_rules_repo, semgrep_rules_repo_dir]
```
kdudka, line 98: Unless you want to checkout a specific snapshot of the repo, you should clone with `--depth=1` to avoid fetching git objects that will not be used.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+class PluginProps:
+    def __init__(self):
+        self.description = (
+            "A fast, open-source, static analysis engine for finding bugs, "
+            "detecting dependency vulnerabilities, and enforcing code standards."
+        )
+
+        # set to True to include this plug-in in `csmock --all-tools`
```
kdudka, line 41: Please update the comment.  It does not make sense to enable the plug-in with `--all-tools` when it requires some extra options.  Plug-ins enabled by `--all-tools` need to work out of the box.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,168 @@
...
+
+        # download semgrep rules
+        def fetch_semgrep_rules_hook(results, props):
+            # target dir where semgrep cli and its dependencies are installed
+            semgrep_lib_dir = os.path.join(results.tmpdir, "semgrep_lib")
+            semgrep_rules_repo_dir = os.path.join(results.tmpdir, "semgrep_rules")
+            try:
+                # make sure the lib/rules directory exists
+                os.makedirs(semgrep_lib_dir, mode=0o755, exist_ok=True)
+                os.makedirs(semgrep_rules_repo_dir, mode=0o755, exist_ok=True)
```
kdudka, line 85: You do not need to create the directory in advance.  `git-clone` will create it if it does not exist.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,168 @@
...
+        # run semgrep scan after successful build
+        props.post_install_hooks += [scan_hook]
+
+        # convert the results into the csdiff's JSON format
+        def filter_hook(results):
+            src = results.dbgdir_raw + SEMGREP_SCAN_OUTPUT
+            if not os.path.exists(src):
+                return 0
+            dst = f"{results.dbgdir_uni}/semgrep-scan-results.json"
+            cmd = f"csgrep '{src}' --mode=json --prepend-path-prefix={SEMGREP_SCAN_DIR}/ > '{dst}'"
```
kdudka, line 165: This looks wrong.  Did you intend to use `--strip-path-prefix` instead?

kdudka (changes requested):
We need to update csdiff to recognize the `semgrep` tool and report the findings as `SEMGREP_WARNING` rather than `UNKNOWN_SARIF_WARNING`.

Please also remove the trailing dot from the git commit title: `... in make-srpm.sh.`

py/plugins/semgrep.py
rhyw: I guess so. I know little about the csdiff options. After updated and I got below records in the output json:
```
    "defects": [
        {
            "checker": "UNKNOWN_SARIF_WARNING",
            "tool": "unknown-sarif",
            "key_event_idx": 0,
            "events": [
                {
                    "file_name": "/var/lib/mock/rhel-8.5.0.z-x86_64/root/builddir/build/BUILD/gzip-1.9/tests/list",
                    "line": 28,
                    "column": 1,
                    "event": "warning[tmp.csmocksyi3zdu7.semgrep_rules.rules.semgrep-rules.bash.lang.best-practice.useless-cat]",
                    "message": "Useless call to 'cat' in a pipeline. Use '<' and '>' for any command to read from a file or write to a file.",
                    "verbosity_level": 0
                }
            ]
        },
...
]
```

Is the path in it correct?

py/plugins/semgrep.py
rhyw: Removed in latest patch.

py/plugins/semgrep.py
rhyw: I removed this comment.

py/plugins/semgrep.py
rhyw: added.

py/plugins/semgrep.py
rhyw: I get it. So I've updated in `handle_args` by a `parser.error` instead. Please review.

rhyw:
I think i've addressed all the issues within this patch. The remaining issue is to update csdiff to recognize "SEMGREP_WARNING" instead of "UNKNOWN_SARIF_WARNING".

I'm not sure if I can do that since it's written in C++ .  I could give it a try later this week.

py/plugins/semgrep.py
kdudka: Obviously not.  The path should be `gzip-1.9/tests/list`.  The paths in the scan results need to be specific to the project being scanned, not to the scanning environment.  Such prefixes would break deduplication of findings and elimination of false positives.  You can run `csmock --all-tools` on the same package to see how the file paths look like with the already integrated scanners.

kdudka:
> I think i've addressed all the issues within this patch. The remaining issue is to update csdiff to recognize "SEMGREP_WARNING" instead of "UNKNOWN_SARIF_WARNING".

You can see how this was done for the Gitleaks tool: https://github.com/csutils/csdiff/commit/54e87362c45d2d774b948bf07298c79ba91a2d54

> I'm not sure if I can do that since it's written in C++ . I could give it a try later this week.

Please ask @hanchuntao for help.  He was already reviewing some csdiff pull requests.

py/plugins/semgrep.py
lzaoral: The `event` key also looks a bit suspicious:
```json
"event": "warning[tmp.csmocksyi3zdu7.semgrep_rules.rules.semgrep-rules.bash.lang.best-practice.useless-cat]",
                  ^^^^^^^^^^^^^^^^^^^
```

rhyw: I'm now working on the csdiff c++ patch to recognize semgrep warnings.

jperezdealgaba:
@rhyw Do we have a reproducer for this? And second, are PyLint findings going to be addressed?

rhyw:
> @rhyw Do we have a reproducer for this? And second, are PyLint findings going to be addressed?

I tested by `csmock -r /etc/mock/rhel-8.5.0.z-x86_64.cfg -t semgrep --semgrep-rules-repo=<the-internal-masssast-repo> --force gzip-1.9-13.el8_5.src.rpm`. However we'll need a renewed `csgrep` package(build from https://github.com/csutils/csdiff/pull/162) to verify the warning recognition.

As for Pylint warning, yeah they're very anonying. I'm curious how other modules survived this. I'll give an update tomorrow, to resolve as much as I can.

py/plugins/semgrep.py
rhyw: MR: https://github.com/csutils/csdiff/pull/162

kdudka: The csdiff merge request is unrelated to the path striping.  The problem mentioned by @lzaoral has not been fixed.   The event name `warning[csmockexonlktd.semgrep_rules.rules.semgrep-rules.bash.lang.correctness.unquoted-variable-expansion-in-command]` should obviously not contain name of the temporary directory, which is different for each run.

rhyw: the csdiff MR was for warning recgonition. the temp rules path was stripped in this MR. see https://github.com/csutils/csmock/pull/149/files#diff-4efa9af9c12deec9697e4bff1d259e971df87c11cbc0a5e99eaee5633f914e78R210

kdudka: I am still getting `warning[csmockc_n6ld1p.semgrep_rules.rules.semgrep-rules.bash.lang.correctness.unquoted-variable-expansion-in-command]` with the latest version of your pull request.  You also need to rebase your changes on top of `main`.

rhyw: The root cause is that if the semgrep scan was run under `/tmp`, which happens to be the same directory with where the rules were located, the semgrep would strip the preceding `/tmp.` automatically in this case. The sed command takes effect when it's run elsewhere other than `tmp.`. 

I've updated the sed with `sed 's|[^[:space:]\"]*{tmp_path}||`, so that it strips the path no matter whether "tmp" is in or not. Please review!

kdudka: As discussed on Slack, the `sed` command does not seem to work as expected.  Please look at `debug/scan-results-all.js` - the values assigned to the `event` fields are obviously invalid.  They contain non-matching `]`.

rhyw: It appears the previous sed has unintendedly stripped the left square bracket. I've simplified the sed by using `sed 's|[^\[]*{tmp_path}||'` since I noticed that the non-whitespace/non-double-quote is not needed, and should be using non-square bracket instead.

I had several tests and it's working as expected. Please take a look. Apologies for the inconveniences it brings.

kdudka (changes requested):

py/plugins/semgrep.py
```diff
@@ -0,0 +1,220 @@
...
+        self.enabled = True
+
+    def init_parser(self, parser):
+        """
+        Initialize the argument parser for the Semgrep plugin.
+        """
+        parser.add_argument(
+            "--semgrep-metrics",
+            default=DEFAULT_SEMGREP_SEND_METRICS,
+            help="configures how usage metrics are sent to the Semgrep server",
```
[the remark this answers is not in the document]
rhyw: This was actually the helper message from `--metrics` doc. Anyway updated.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,220 @@
...
+
+        parser.add_argument(
+            "--semgrep-rules-repo",
+            help="semgrep rules repo, assuming rules are located under the 'rules' sub-directory",
+        )
+
+        parser.add_argument(
+            "--semgrep-verbose",
+            default=DEFAULT_SEMGREP_VERBOSE,
+            help="show more details about what rules are running, which files failed to parse, etc",
```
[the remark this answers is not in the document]
rhyw: providing default value is indeed confusing. I've updated to use `store_true` instead, and removed the redundant default.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,220 @@
...
+                results.error("failed to install semgrep cli using pip")
+
+            semgrep_prefix = f"env PATH={semgrep_lib_dir}/bin:$PATH, PYTHONPATH={semgrep_lib_dir}"
+
+            semgrep_rules_repo_dir = os.path.join(results.tmpdir, "semgrep_rules")
+            repo_clone_cmd = [
+                "git",
+                "clone",
+                "--depth",
+                "1",
```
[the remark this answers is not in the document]
rhyw: updated

py/plugins/semgrep.py
```diff
@@ -0,0 +1,220 @@
...
+            os.remove(f"{results.dbgdir_raw}{SEMGREP_SCAN_CHROOT_ROOT_PATH}")
+
+            tmp_dir_basename = results.tmpdir.split("/")[-1]
+            semgrep_rules_path_prefix = f"{tmp_dir_basename}/semgrep_rules/"
+            # semgrep report has dot-separated rules path
+            tmp_path = semgrep_rules_path_prefix.lstrip("/").replace("/", r"\.")
+            # strip suspicious path prefix from the semgrep rules directory
+            # depending on where the semgrep scan process is run, the raw report may or may not contain "/tmp"
+            # in its rules path. The following sed command strips suspicious path prefixes by removing
+            # any sequence of non left-square-bracket charactors preceding '{tmp_path}'
```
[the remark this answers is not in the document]
rhyw: updated.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,220 @@
+# Copyright (C) 2014 Red Hat, Inc.
```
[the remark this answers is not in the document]
rhyw: updated.

py/plugins/semgrep.py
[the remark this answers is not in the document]
kdudka: I have further tweaked this in d1c58c4812556f653e7edad67d712fafe5b7d6fd.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,215 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with csmock.  If not, see <http://www.gnu.org/licenses/>.
+
+import os
+
+# disable metrics to be sent to semgrep cloud
+DEFAULT_SEMGREP_SEND_METRICS = "off"
+
+SEMGREP_CLI_VERSION = "1.56.0"
```
kdudka, line 25: Do we want to use version 1.56.0 specifically?  The current latest release is 1.63.0.  At minimum, we should put here a comment explaining where to check for a more recent version of Semgrep, as proposed in a2ecae5360a52bb0aec94cacb64957f8f7e4198d.

rhyw: the rules used is incompatible with a newer cli: https://gitlab.cee.redhat.com/prodsec-services/masast/-/issues/1

I think tests are needed against the rules for newer cli release unless rules are generic enough.

rhyw:
Any more updates needed please?

py/plugins/semgrep.py
```diff
@@ -0,0 +1,218 @@
...
+    def handle_args(self, parser, args, props):
+        if not self.enabled:
+            return
+
+        if not args.semgrep_rules_repo:
+            parser.error("'--semgrep-rules-repo' is required to run semgrep scan")
+            return 1
+
+        # download semgrep rules
+        def fetch_semgrep_rules_hook(results, props):
```
kdudka, line 92: The comment and name of the function are misleading because the function not only downloads the rules but mainly installs the semgrep tool itself.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,218 @@
...
+
+        props.pre_mock_hooks += [fetch_semgrep_rules_hook]
+
+        def scan_hook(results, mock, props):
+            semgrep_lib_dir = os.path.join(results.tmpdir, "semgrep_lib")
+            semgrep_prefix = f"env PATH={semgrep_lib_dir}/bin:$PATH PYTHONPATH={semgrep_lib_dir}"
+            # assuming semgrep rules are located under the 'rules' directory
+            semgrep_rules_dir = os.path.join(results.tmpdir, "semgrep_rules/rules")
+            # get the chroot root path
+            ec, output = results.get_cmd_output(mock.get_mock_cmd(["--print-root-path"]), shell=False)
```
kdudka, line 147: This step should not be needed any more.  You can get the information from `mock.mock_root`, thanks to #151.

py/plugins/semgrep.py
rhyw: that's indeed confusing. addressed in 2f274c484674de415abbdcb6de7bc9175ddff178

py/plugins/semgrep.py
rhyw: Thanks, addressed in a79095cc48c897d858e753164100c3034a1c904a

kdudka:
@rhyw Did you look at the warnings produced by Differential PyLint?  For example, this one looks valid:
```
Error: PYLINT_WARNING:
py/plugins/semgrep.py:83:4: R1710[inconsistent-return-statements]: Plugin.handle_args: Either all return statements in a function should return an expression, or none of them should.
```

As far as I can see, no other csmock plug-in returns any value from `handle_args()`.  I find it confusing that the `semgrep` plug-in (in some cases) does.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,212 @@
...
+            semgrep_scan_cmd = semgrep_prefix + (
+                f" semgrep scan --metrics={args.semgrep_metrics} --sarif"
+                f" --config={semgrep_rules_dir}"
+            )
+            if args.semgrep_verbose:
+                semgrep_scan_cmd += " --verbose"
+
+            # append additional options passed to the 'semgrep scan' command
+            if args.semgrep_scan_opts:
+                semgrep_scan_cmd += f" {args.semgrep_scan_opts}"
```
kdudka, line 159: This introduces a shell injection vulnerability as described at [CVE-2024-2243](https://bugzilla.redhat.com/CVE-2024-2243) but with higher impact because this command does not run in the isolated `mock` environment.  Please use `sanitize_opts_arg()` from `csmock.common.util` to sanitize the input.

kdudka (changes requested):

py/plugins/semgrep.py
rhyw: addressed in 8c4dbfb10169487c59809730fe71415a393794d8

@kdudka However I think `sanitize_opts_arg()` needs update to work with non-snyk opts, specifically https://github.com/rhyw/csmock/blob/main/py/common/util.py#L53

rhyw: Fix attempt: https://github.com/csutils/csmock/pull/156

kdudka: Sorry for the copy/paste error.  The commit in in question did not go through the usual upstream review process because of the embargo and it somehow slipped through the internal review.

lzaoral:
@rhyw It is possible to execute the semgrep plugin without any need for additional resources?  If yes, could you please a test description for it?  You can take a look at the one for the GCC plugin as the inspiration: https://github.com/csutils/csmock/commit/219e34ee4daeb0be0028f50ce194e917b1de4dce

kdudka:
@lzaoral If we want to extend test coverage, I would focus on plug-ins with `self.stable = True` in `PluginProps`.  All other plug-ins are experimental.  This plug-in downloads the full software stack of `semgrep` over network, which is going to be slow and prone to fail for gazillion of reasons unrelated to the changes of `csmock` code.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,227 @@
...
+            # run semgrep scan
+            ec = results.exec_cmd(semgrep_scan_cmd, shell=True)
+
+            # according to semgrep cli scan doc, below are all possible return codes
+            if ec == 123:
+                results.error("semgrep: Indiscriminate errors reported on standard error.")
+            elif ec == 124:
+                results.error("semgrep: Command line parsing errors.")
+            elif ec == 125:
+                results.error("semgrep: Unexpected internal errors (bugs).")
```
kdudka, line 190: The error handling is wrong.  There needs to be a catch-all `else` branch to handle any error code that is not special-cased above.  I tried to run `csmock` like this:
```
csmock -r fedora-40-x86_64 -t semgrep -f /home/kdudka/git/csmock/csmock-3.5.3.20240321.201842.g4014c4e.pr_149-1.fc39.src.rpm --semgrep-scan-opts='--help; cat /home/kdudka/.config/configstore/snyk.json #' --semgrep-rules-repo=https://gitlab.cee.redhat.com/prodsec-services/masssast.git
```
... and it finished successfully despite `semgrep-scan.log` contained:
```
semgrep scan: unknown option '--help;', did you mean '--help'?
Usage: semgrep scan [OPTION]… [TARGETS]…
Try 'semgrep scan --help' for more information.
Error: fatal error
Exiting with error status 2: osemgrep scan --metrics=off --sarif --config=/tmp/csmockngt2zd0y/semgrep_rules/rules --help; cat /home/kdudka/.config/configstore/snyk.json # --output=/var/lib/mock/fedora-40-x86_64/root//builddir/semgrep-scan-results.sarif /var/lib/mock/fedora-40-x86_64/root//builddir/build/BUILD
```

kdudka (changes requested):

py/plugins/semgrep.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+class Plugin:
+    """
+    Semgrep static analysis engine plugin
+    """
+    def __init__(self):
+        self.enabled = False
+
+    def get_props(self):  # pylint: disable=missing-function-docstring
```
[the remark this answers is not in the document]
rhyw: without this, pylint raises a warning.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def handle_args(self, parser, args, props):  # pylint: disable=too-many-statements,missing-function-docstring
+        if not self.enabled:
+            return
+
+        if not args.semgrep_rules_repo:
+            parser.error("'--semgrep-rules-repo' is required to run semgrep scan")
+
+        # sanitize options passed to --semgrep-scan-opts to avoid shell injection
+        self.semgrep_scan_opts = sanitize_opts_arg(parser, args, "--semgrep-scan-opts")
```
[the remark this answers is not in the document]
rhyw: This is to be consistent with other plugins setup. moving it to `__init__` requires updates in the callee(py/csmock::main), which may require updates of other plugins too.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,227 @@
...
+                os.makedirs(semgrep_lib_dir, mode=0o755, exist_ok=True)
+            except OSError:
+                results.error("failed to create semgrep lib directory")
+                return 1
+
+            # install semgrep cli using pip
+            cmd = f"python3 -m pip install --target={semgrep_lib_dir} semgrep=={SEMGREP_CLI_VERSION}"
+            ec = results.exec_cmd(cmd, shell=True)
+            if 0 != ec:
+                results.error("failed to install semgrep cli using pip")
```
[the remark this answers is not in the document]
rhyw: good catch, I'll update to return values consistently.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,227 @@
...
+                return ec
+            # query version of semgrep
+            cmd = semgrep_prefix + " semgrep --version"
+            ec, output = results.get_cmd_output(cmd)
+            if 0 != ec:
+                results.error("failed to query semgrep cli version", ec=ec)
+                return ec
+
+            # parse and record the version of semgrep cli
+            version = output.rstrip("\n")
```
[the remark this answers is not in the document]
rhyw: both should work, but I think `rstrip` is better because there is no leading `\n` in the output.

py/plugins/semgrep.py
```diff
@@ -0,0 +1,227 @@
...
+        if not args.semgrep_rules_repo:
+            parser.error("'--semgrep-rules-repo' is required to run semgrep scan")
+
+        # sanitize options passed to --semgrep-scan-opts to avoid shell injection
+        self.semgrep_scan_opts = sanitize_opts_arg(parser, args, "--semgrep-scan-opts")
+
+        # install semgrep cli and download semgrep rules
+        def prepare_semgrep_runtime_hook(results, props):
+            # target dir where semgrep cli and its dependencies are installed
+            semgrep_lib_dir = os.path.join(results.tmpdir, "semgrep_lib")
```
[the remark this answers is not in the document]
rhyw: Although this should also work, the top level `semgrep` needs to exist. The rules repo is specified by user and a repo clone would create the rules directory automatically, putting the rules inside the `semgrep` adds a pre-requisite.

py/plugins/semgrep.py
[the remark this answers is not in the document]
rhyw: both should work, but I think `rstrip` is better because there is no leading `\n` in the output.

py/plugins/semgrep.py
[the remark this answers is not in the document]
rhyw: I mis-interpreted this. Thanks, addressed in 1d0a24ff1dd6f4a007c95291fe83ecd9ee36ca30

py/plugins/semgrep.py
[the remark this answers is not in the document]
rhyw: addressed in 66f0ce1a0dba5fa73725199c49af13ed306cba0e

py/plugins/semgrep.py
rhyw: addressed in 66f0ce1a0dba5fa73725199c49af13ed306cba0e 

had some tests, if unrecognized opts were specified, it should raise error as below:
```
!!! 2024-03-27 04:54:01 error: semgrep: Scan failed with exit code 2. Command: env PATH=/tmp/csmocktczvra25/semgrep_lib/bin:$PATH PYTHONPATH=/tmp/csmocktczvra25/semgrep_lib semgrep scan --metrics=off --sarif --config=/tmp/csmocktczvra25/semgrep_rules/rules '--help;' cat /home/kdudka/.config/configstore/snyk.json '#' --output=/var/lib/mock/rhel-9-x86_64/root//builddir/semgrep-scan-results.sarif /var/lib/mock/rhel-9-x86_64/root//builddir/build/BUILD 2>/var/lib/mock/rhel-9-x86_64/root//builddir/semgrep-scan.log

>>> 2024-03-27 04:54:01 "'/usr/bin/mock' '-r' '/etc/mock/rhel-9-x86_64.cfg' '--plugin-option=tmpfs:keep_mounted=True' '--config-opts=print_main_output=True' '--quiet' '--shell' 'tar -c --remove-files '/builddir/semgrep-scan-results.sarif' '/builddir/semgrep-scan.log' '/builddir/semgrep-chroot-root'' | tar -xC '/tmp/csmocktczvra25/gzip-1.10-9.el9_0/debug/raw-results'"
tar: Removing leading `/' from member names
tar: /builddir/semgrep-scan-results.sarif: Cannot stat: No such file or directory
tar: Removing leading `/' from hard link targets
tar: Exiting with failure status due to previous errors

>>> 2024-03-27 04:54:02 "/usr/bin/mock" "-r" "/etc/mock/rhel-9-x86_64.cfg" "--plugin-option=tmpfs:keep_mounted=True" "--config-opts=print_main_output=True" "--quiet" "--clean"

scan.ini: time-finished = 2024-03-27 04:54:04
scan.ini: exit-code = 1
<<< 2024-03-27 04:54:04 csmock exit code: 1
```

kdudka:
@rhyw Looks good.  Please squash the fixup commits.  Namely 4014c4e1b1d52993974feb9f39dfcb418cde5811 needs to be squashed before merging so that we do not unnecessarily introduce and fix a security issue in the git history.

rhyw:
> @rhyw Looks good. Please squash the fixup commits. Namely [4014c4e](https://github.com/csutils/csmock/commit/4014c4e1b1d52993974feb9f39dfcb418cde5811) needs to be squashed before merging so that we do not unnecessarily introduce and fix a security issue in the git history.

Thanks, commits squashed

kdudka:
@rhyw Thanks!  I have squashed my fixup commits, too.  We do not want them in the `main` branch.

kdudka (approved):

PR #150, opened 2024-01-26
results: skip comments in exclude-paths.txt

stefan-cornelius:
This allows comments in the path exlusion regex file. Instead of adding each line as a filtering rule, it skips empty lines and lines starting with a # symbol.

Fixes https://issues.redhat.com/browse/OSH-508

### DIFF
--- a/py/common/results.py
+++ b/py/common/results.py
@@ -404,5 +404,8 @@ def handle_known_fp_list(props, results):
         lines = file_handle.readlines()
         for line in lines:
             path_re = line.strip()
+            if len(path_re) == 0 or path_re.startswith("#"):
+                # skip comments and empty lines
+                continue
             filter_cmd = f'csgrep --mode=json --invert-match --path="{shell_quote(path_re)}"'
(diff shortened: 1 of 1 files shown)


stefan-cornelius:
Just noticed the comment about the sanity check in OSH-508. Will add ...

kdudka:
@stefan-cornelius The code change looks good.  Please delimit the commit title with a blank line as suggested in the `git-commit(1)` man page:
```
Though not required, it’s a good idea to begin the commit message with a single short (no more than 50 characters)
line summarizing the change, followed by a blank line and then a more thorough description. The text up to the first
blank line in a commit message is treated as the commit title, and that title is used throughout Git. For example,
git-format-patch(1) turns a commit into email, and it uses the title on the Subject line and the rest of the commit
in the body.
```
I propose this as the commit title: `results: skip comments in exclude-paths.txt`

kdudka (approved):

kdudka:
@stefan-cornelius Merging, thanks!

PR #151, opened 2024-02-19
csmock: use the root mock profile property for lock names

lzaoral:
To fix a crash when two mock profile configuration files have different names but use the same root directory:
```
[...]
>>> 2024-02-19 14:29:12	"/usr/bin/mock" "-r" "/tmp/tmpyfvruenl/mock.cfg" "--plugin-option=tmpfs:keep_mounted=True" "--config-opts=print_main_output=True" "--init"
INFO: mock.py version 5.2 starting (python version = 3.9.18, NVR = mock-5.2-1.kdudka.5.el9), args: /usr/libexec/mock/mock -r /tmp/tmpyfvruenl/mock.cfg --plugin-option=tmpfs:keep_mounted=True --config-opts=print_main_output=True --init
Start(bootstrap): init plugins
INFO: selinux enabled
Finish(bootstrap): init plugins
Start: init plugins
INFO: selinux enabled
Finish: init plugins
INFO: Signal handler active
Start: run
Start: clean chroot
ERROR: Build root is locked by another process.

!!! 2024-02-19 14:29:13	warning: failed to init mock profile (/tmp/tmpyfvruenl/mock.cfg), trying to scrub cache...
[...]
```

Resolves: https://github.com/openscanhub/openscanhub/pull/228#discussion_r1494645666

### DIFF
--- a/py/csmock
+++ b/py/csmock
@@ -22,6 +22,7 @@ import argparse
 import copy
 import importlib
 import os
+import pathlib
 import pkgutil
 import re
 import shlex
@@ -193,10 +194,6 @@ class MockWrapper:
         self.results = results
         self.mock_profile = props.mock_profile
         self.mock_root_override = props.mock_root_override
-        lock = f"/tmp/.csmock-" + (self.mock_profile if not self.mock_root_override else
-                                   self.mock_root_override).replace('/','_')
-        self.lock_file = f"{lock}.lock"
-        self.meta_lock_file = f"{lock}.metalock"
         self.pid = os.getpid()
         self.scrub_done = props.skip_mock_init
         self.init_done = props.skip_mock_init
@@ -206,6 +203,24 @@ class MockWrapper:
         self.add_repos = props.add_repos
         # just to silence pylint, will be initialized in __enter__()
         self.def_cmd = None
+
+        # get buildroot directory
+        lock_name = self.mock_root = self.mock_root_override
+        if not self.mock_root:
+            cmd = ['mock', '-r', self.mock_profile, '--print-root-path']
+            ec, self.mock_root = results.get_cmd_output(cmd, shell=False)
+            if ec != 0:
+                results.error(f'mock could not determine root path for {self.mock_profile}', ec=ec)
+
+            # strip trailing newline
+            self.mock_root = self.mock_root.strip()
+
+            # use only the basename of the mock root
+            lock_name = pathlib.Path(self.mock_root).parent.stem
+
+        lock = f"/tmp/.csmock-{lock_name.replace('/', '_')}"
+        self.lock_file = f"{lock}.lock"
+        self.meta_lock_file = f"{lock}.metalock"
 
     def __enter__(self):
         cmd = "flock -w%u '%s' -c '\


py/csmock
```diff
@@ -206,6 +202,17 @@ class MockWrapper:
         # just to silence pylint, will be initialized in __enter__()
         self.def_cmd = None
 
+        # get buildroot directory
+        ec, buildroot_dir = results.get_cmd_output(['mock', '-r', self.mock_profile, '--print-root-path'],
```
kdudka, line 206: You should remove the trailing newline from the output.  Otherwise it looks like this:
```
[...]
>>> 2024-02-20 09:06:04 "rpm -qa | sort -V > '/tmp/csmockdjcpcck3/csmock-3.5.1-1.fc41/debug/rpm-list-host.txt'"

>>> 2024-02-20 09:06:06 tip: you can use --root-override=<directory> to run this csmock instance in parallel
>>> 2024-02-20 09:06:06 waiting till /tmp/.csmock-_var_lib_mock_rhel-9-x86_64_root_
.lock (PID 2508) disappears...
>>> 2024-02-20 09:07:06 waiting till /tmp/.csmock-_var_lib_mock_rhel-9-x86_64_root_
.lock (PID 2508) disappears...
```

py/csmock
```diff
@@ -206,6 +202,17 @@ class MockWrapper:
...
+                                                   shell=False)
+        if ec != 0:
+            results.error(f'mock could not determine root path for {self.mock_profile}', ec=0)
+            buildroot_dir = self.mock_profile
+
+        lock = f"/tmp/.csmock-" + (self.mock_root_override or buildroot_dir).replace('/','_')
```
kdudka, line 212: You only use `buildroot_dir` if `mock_root_override` is None.  If `mock_root_override` is set, it does not make much sense to execute `--print-root-path` first.

kdudka (changes requested):
Would it make sense to cache the `root_path` value in the `MockWrapper` object so that it can be used at other places (e.g. in the `semgrep` plug-in) without querying it again?

py/csmock
lzaoral: You are right, but due to https://github.com/csutils/csmock/pull/151#pullrequestreview-1889723365 it is going to be executed always.

lzaoral:
All suggestions addressed in or obsoleted by [87dfa62](https://github.com/csutils/csmock/pull/151/commits/87dfa62612854f946ba56080adb7007ef5bf7014).

py/csmock
kdudka: @lzaoral Why?  If `props.mock_root_override` is set when `MockWrapper` is constructed, we should cache its value.  The value specified in mock config does not really matter if we are going to override it in each `mock` command executed through the `MockWrapper` object.

py/csmock
```diff
@@ -206,6 +202,17 @@ class MockWrapper:
         # just to silence pylint, will be initialized in __enter__()
         self.def_cmd = None
 
+        # get buildroot directory
+        ec, self.mock_root = results.get_cmd_output(['mock', '-r', self.mock_profile, '--print-root-path'],
```
kdudka, line 206: I find this difficult to read.  Could we perhaps format it like this?
```py
cmd = ['mock', '-r', self.mock_profile, '--print-root-path']
ec, self.mock_root = results.get_cmd_output(cmd, shell=False)
```

py/csmock
lzaoral: Got it!  Addressed in 9b24649def5973cc02bda1404cd3233f5de9f246.

py/csmock
lzaoral: Addressed in 9b24649def5973cc02bda1404cd3233f5de9f246.

py/csmock
```diff
@@ -206,6 +202,19 @@ class MockWrapper:
...
+        # get buildroot directory
+        self.mock_root = self.mock_root_override
+        if not self.mock_root:
+            cmd = ['mock', '-r', self.mock_profile, '--print-root-path']
+            ec, self.mock_root = results.get_cmd_output(cmd, shell=False)
+            if ec != 0:
+                results.error(f'mock could not determine root path for {self.mock_profile}', ec=0)
+                self.mock_root = self.mock_profile
+
+        lock = f"/tmp/.csmock-" + self.mock_root.replace('/','_').strip()
```
kdudka, line 214: This makes it incompatible with how `csmock` works today.  To stay compatible, we should take only the base name of the `root` directory.  Otherwise `csmock` will start to use a new lock name after update and could clash with the already running instances.

lzaoral: Addressed in 5a0d5c1ae10cad0fa53e7bb85d6c93b8e187c766.

py/csmock
```diff
@@ -207,6 +203,25 @@ class MockWrapper:
...
+            if ec != 0:
+                results.error(f'mock could not determine root path for {self.mock_profile}', ec=0)
+                self.mock_root = self.mock_profile
+
+            # strip trailing newline
+            self.mock_root = self.mock_root.strip()
+
+        # prefer whole value of self.mock_root_override which was explicitly
+        # provided by the user, otherwise, use only the basename of the mock root
+        lock_name = (self.mock_root_override or os.path.basename(self.mock_root)).replace('/','_')
```
kdudka, line 220: Sorry, my suggestion to use `basename()` was totally off.

`os.path.basename("/var/lib/mock/rhel-9-x86_64/root/")` returns an empty string whereas we need to get `rhel-9-x86_64` out of it (without the `/root/` suffix).

PR #153, opened 2024-02-20
snyk: added snyk stats to metadata

jperezdealgaba:
Related: https://issues.redhat.com/browse/OSH-347
Reproducer: csmock -t snyk --force -r rhel-8-x86_64 osbuild-106-1.el10+4.src.rpm

Added the stats from snyk results (snyk coverage rate, analyzed files and total of files) to the metadata file.

### DIFF
--- a/py/common/snyk.py
+++ b/py/common/snyk.py
@@ -1,0 +1,52 @@
+# Copyright (C) 2024 Red Hat, Inc.
+#
+# This file is part of csmock.
+#
+# csmock is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# any later version.
+#
+# csmock is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with csmock.  If not, see <http://www.gnu.org/licenses/>.
+
+import json
+
+
+def snyk_write_analysis_meta(results, raw_results_file):
+    """write snyk stats on metadata file. At the time, we write the total number of files,
+    the number of supported files and the coverage ratio."""
+
+    try:
+        with open(raw_results_file) as snyk_results_file:
+            data = json.load(snyk_results_file)
+            coverage_stats = data["runs"][0]["properties"]["coverage"]
+            total_files = 0
+            supported_files = 0
+            for lang in coverage_stats:
+                total_files += lang["files"]
+                if lang["type"] == "SUPPORTED":
+                    supported_files += lang["files"]
+
+            coverage_ratio = 0
+            if total_files > 0:
+                coverage_ratio = int(supported_files * 100 / total_files)
+
+            results.ini_writer.append("snyk-scanned-files-coverage", coverage_ratio)
+            results.ini_writer.append("snyk-scanned-files-success", supported_files)
+            results.ini_writer.append("snyk-scanned-files-total", total_files)
+
+            return 0
+
+    except OSError as e:
+        results.error(f"snyk-scan: failed to read {raw_results_file}: {e}")
+        return 1
+
+    except KeyError as e:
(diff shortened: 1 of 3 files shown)


py/plugins/snyk.py
```diff
@@ -204,4 +206,8 @@ def filter_hook(results):
             cmd = FILTER_CMD % (src, dst)
             return results.exec_cmd(cmd, shell=True)
 
-        props.post_process_hooks += [filter_hook]
+        def write_snyk_stats_metadata(results):
+            results_file = results.dbgdir_raw + SNYK_OUTPUT
+            return snyk_write_analysis_meta(results, results_file)
```
jperezdealgaba, line 211: Note: In the future, the `results_file` variable should be always the same path, but until the chroot/snyk problem is not solved, we need to have two different paths.

kdudka (changes requested):
The newly introduced module needs to be installed by CMake to keep the RPM packaging (or local installations) working: https://github.com/csutils/csmock/blob/9befcc342fdf58f28accadd9fd5a7526059a423f/py/CMakeLists.txt#L23

kdudka:
> The newly introduced module needs to be installed by CMake...

@jperezdealgaba Note that this problem had already been revealed by the failing CI tests.

lzaoral (changes requested):

py/plugins/snyk.py
```diff
@@ -17,6 +17,8 @@
 
 import os
 
+from csmock.common.snyk import snyk_write_analysis_meta
```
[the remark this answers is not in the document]
jperezdealgaba: @kdudka I have added the file to the CMake file. 
I have built the rpms locally and the new files are included in the src.rpms. Is something else needed?

py/common/snyk.py
```diff
@@ -0,0 +1,36 @@
+# standard imports
```
kdudka, line 1: Please include the license header as we do in other files in `py/common`.

py/common/snyk.py
```diff
@@ -0,0 +1,36 @@
...
+import json
+
+
+def snyk_write_analysis_meta(results, raw_results_file):
+    """write snyk stats on metatadata file. At the time, we write the total number of files,
```
kdudka, line 6: `metatadata` -> `metadata`

py/common/snyk.py
```diff
@@ -0,0 +1,36 @@
...
+                coverage_ratio = int(supported_files * 100 / total_files)
+
+            results.ini_writer.append("snyk-scanned-files-coverage", coverage_ratio)
+            results.ini_writer.append("snyk-scanned-files-success", supported_files)
+            results.ini_writer.append("snyk-scanned-files-total", total_files)
+
+            return 0
+
+    except OSError as e:
+        results.error(f"snyk-scan: snyk-results.sarif file not found: {e}")
```
kdudka, line 31: `OSError` can be returned also in other cases, such as insufficient permissions to access the file, so the error message should be more generic, e.g.: `... failed to read {raw_results_file}: {e}`

kdudka (changes requested):

py/common/snyk.py
```diff
@@ -0,0 +1,52 @@
+# Copyright (C) 2024 Red Hat, Inc.
```
jperezdealgaba, line 1: @kdudka I have updated the license header to refer to 2024. Should I do another commit to update plugin/snyk header?

kdudka: @jperezdealgaba Thanks!  I do not think it is critical.  We may eventually migrate `csutils` to use the two-lines SPDX license headers like we did for OSH: https://github.com/openscanhub/openscanhub/commit/31156512208145d032c470018cb338bb219e2059

@lzaoral What do you think about it?

lzaoral: @kdudka It is not necessary.

kdudka: @lzaoral Sure.  I was just thinking about modernization of `csutils` so that we do not waste time on maintaining the legacy GPL headers in source code files.

lzaoral: @kdudka I was reacting to @jperezdealgaba's idea.

Of course, I agree with modernisation of licensing information.

kdudka (approved):
Looks good.  Thanks for the update!

PR #154, opened 2024-02-20
csmock: add `--scrub-on-exit` option

lzaoral:
... to pass `--scrub=all` to the underlying `mock` instance.

Related: https://github.com/openscanhub/openscanhub/pull/228#discussion_r1494707783

### DIFF
--- a/py/csmock
+++ b/py/csmock
@@ -200,6 +200,7 @@ class MockWrapper:
         self.pid = os.getpid()
         self.scrub_done = props.skip_mock_init
         self.init_done = props.skip_mock_init
+        self.scrub_on_exit = props.scrub_on_exit
         self.skip_clean = props.skip_mock_clean
         self.use_login_shell = props.use_login_shell
         self.add_repos = props.add_repos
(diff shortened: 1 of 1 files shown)


lzaoral:
@praiskup It seems that differential PyLint reports a problem in a piece of code that was edited in this PR. 🤔

py/csmock
```diff
@@ -782,6 +789,10 @@ exceeds the specified limit (defaults to 1024).")
         "--skip-build", action="store_true",
         help="do not run %%build and %%install sections [EXPERIMENTAL]")
 
+    parser.add_argument(
+        "--scrub", action="store_true",
```
kdudka, line 793: I would make the option name more specific, such as `--scrub-on-exit` or `--scrub-after-scan`.

kdudka (changes requested):
Should we handle it better when someone uses both `--scrub` and `--no-clean`?

py/csmock
lzaoral: Addressed in 3b0f6953f2c429d93d8d98f3d44eddfdcd5a9644.

PR #156, opened 2024-03-21
util.py: improve sanitize_opts_arg to be reused by other plug-ins

rhyw:


py/common/util.py
```diff
@@ -42,15 +42,29 @@ def arg_value_by_name(parser, args, arg_name):
...
         return None
 
+    dest = convert_option_to_dest(parser, arg_name)
+    if dest is None:
+        return None
+
     # split, quote, and rejoin the options to avoid shell injection
     try:
-        split_opts = shlex.split(args.snyk_code_test_opts)
+        split_opts = shlex.split(getattr(args, dest))
```
kdudka, line 67: This is a copy/paste error.  It should have been `shlex.split(opts_str)` obviously.  I think all the other code changes are redundant.

kdudka (changes requested):

py/common/util.py
rhyw: Updated per suggested, had a quick test, working as expected. Thanks for suggestion

kdudka (approved):
Looks good.  Thanks for the update!

rhyw:
Thanks

PR #157, opened 2024-03-26
rust-clippy: add rust clippy plugin

petr-matousek:
Resolves: https://issues.redhat.com/browse/OSH-30


make-srpm.sh
```diff
@@ -252,6 +255,13 @@ Requires: csmock-common
 %description -n csmock-plugin-unicontrol
 This package contains the unicontrol plug-in for csmock.
 
+%package -n csmock-plugin-clippy
```
kdudka, line 258: Please keep the sub-packages sorted alphabetically.

make-srpm.sh
```diff
@@ -359,6 +369,12 @@ This package contains the unicontrol plug-in for csmock.
 %{_datadir}/csmock/scripts/find-unicode-control.py*
 %{python3_sitelib}/csmock/plugins/unicontrol.py*
 %{python3_sitelib}/csmock/plugins/__pycache__/unicontrol.*
+
+%files -n csmock-plugin-clippy
```
kdudka, line 373: Please keep the sub-packages sorted alphabetically.

make-srpm.sh
```diff
@@ -359,6 +369,12 @@ This package contains the unicontrol plug-in for csmock.
...
+%{_datadir}/csmock/scripts/clippy.sh
```
kdudka, line 374: I would rename the script to `inject-clippy.sh` to make it obvious what the script does.

py/CMakeLists.txt
```diff
@@ -71,4 +71,5 @@ if(ENABLE_CSMOCK)
     install_plugin(symbiotic)
     install_plugin(valgrind)
     install_plugin(unicontrol)
+    install_plugin(clippy)
```
kdudka, line 74: Please keep the list sorted.

py/plugins/clippy.py
```diff
@@ -0,0 +1,44 @@
+import csmock.common.util
+
+RUN_CLIPPY_CONVERT = "python3 /usr/share/csmock/scripts/convert-clippy.py"
```
kdudka, line 3: Do we need to invoke the Python interpreter explicitly?

scripts/clippy.sh
```diff
@@ -0,0 +1,28 @@
...
+fi
+
+mv $ORIGINAL_BINARY $NEW_LOCATION
+
+cat > $ORIGINAL_BINARY << 'EOF'
+#!/bin/bash
+
+ORIGINAL_PARAMS=("$@")
+
+if [[ $1 == "build" ]]; then
```
kdudka, line 18: Is it safe to expect `build` to always be the first argument of `cargo`?  The help output of `cargo` says:
```
Usage: cargo [OPTIONS] [COMMAND]
```
At minimum, I would insert a FIXME comment here.

scripts/clippy.sh
```diff
@@ -0,0 +1,28 @@
...
+    set -- "clippy" "${@:2}"
+    REPLACE_ME "$@" --message-format=json >> /builddir/clippy-output.txt
+fi
+
+REPLACE_ME "${ORIGINAL_PARAMS[@]}"
+EOF
+
+sed -i "s|REPLACE_ME|$NEW_LOCATION|g" $ORIGINAL_BINARY
```
kdudka, line 26: I would combine it with the previous command so that we do not write the script twice to the file system:
```diff
--- a/scripts/clippy.sh
+++ b/scripts/clippy.sh
@@ -7,22 +7,20 @@ if [ -f "$NEW_LOCATION" ]; then
     rm "/builddir/clippy-output.txt"
     exit 0
 fi

 mv $ORIGINAL_BINARY $NEW_LOCATION

-cat > $ORIGINAL_BINARY << 'EOF'
+sed "s|REPLACE_ME|$NEW_LOCATION|g" > $ORIGINAL_BINARY << 'EOF'
 #!/bin/bash

 ORIGINAL_PARAMS=("$@")

 if [[ $1 == "build" ]]; then
     set -- "clippy" "${@:2}"
     REPLACE_ME "$@" --message-format=json >> /builddir/clippy-output.txt
 fi

 REPLACE_ME "${ORIGINAL_PARAMS[@]}"
 EOF

-sed -i "s|REPLACE_ME|$NEW_LOCATION|g" $ORIGINAL_BINARY
-
 chmod +x $ORIGINAL_BINARY
```

scripts/convert-clippy.py
```diff
@@ -0,0 +1,42 @@
...
+import sys
+import re
+
+
+def main():
+    pattern = r"^(.*?): (.*?)\s+-->\s+(.*?):(\d+):\d+(.*)"
+    pattern = re.compile(pattern, re.DOTALL)
+
+    for line in sys.stdin:
+        item = json.loads(line)
```
kdudka, line 13: Should we catch an exception if the current line fails to parse and continue with the next line?  As I understand it, a single illegal byte would terminate the script and skip all the input that has not been read yet.

scripts/convert-clippy.py
```diff
@@ -0,0 +1,42 @@
...
+
+        if item["reason"] != "compiler-message":
+            continue
+
+        match = re.search(r"path\+file://(.*)\)", item["package_id"])
```
kdudka, line 18: I would appreciate some comments explaining why/how we handle `package` here.

py/plugins/clippy.py
```diff
@@ -0,0 +1,44 @@
...
+        self.enabled = False
+
+    def get_props(self):
+        return PluginProps()
+
+    def enable(self):
+        self.enabled = True
+
+    def init_parser(self, parser):
+        csmock.common.util.install_script_scan_opts(parser, "clippy")
```
kdudka, line 23: Unless I am overlooking something, this will create command-line options that will not have any effect.  I think it should be removed.  This call is usually paired with `dirs_to_scan_by_args()`: https://github.com/csutils/csmock/blob/87d976186fefd36278bfa66f68ee0c2f53ba4079/py/plugins/shellcheck.py#L55

scripts/clippy.sh
```diff
@@ -0,0 +1,28 @@
...
+mv $ORIGINAL_BINARY $NEW_LOCATION
+
+cat > $ORIGINAL_BINARY << 'EOF'
+#!/bin/bash
+
+ORIGINAL_PARAMS=("$@")
+
+if [[ $1 == "build" ]]; then
+    set -- "clippy" "${@:2}"
+    REPLACE_ME "$@" --message-format=json >> /builddir/clippy-output.txt
```
kdudka, line 20: Would it make sense to insert `set -x` before this command to get the actual scanning command(s) recorded in `scan.log` for debugging/auditing purposes?

py/plugins/clippy.py
```diff
@@ -0,0 +1,44 @@
...
+
+        props.install_pkgs += ["clippy"]
+        props.copy_out_files += [CLIPPY_OUTPUT]
+
+        csmock.common.util.install_default_toolver_hook(props, "clippy")
+
+        def convert_hook(results):
+            src = f"{results.dbgdir_raw}{CLIPPY_OUTPUT}"
+            dst = f"{results.dbgdir_uni}/clippy-capture.err"
+            cmd = f'{RUN_CLIPPY_CONVERT} < {src} | csgrep --prepend-path-prefix="/builddir/build/BUILD/" --remove-duplicates > {dst}'
```
kdudka, line 41: Could we optimize out the removal and restoration of the `/builddir/build/BUILD/` prefix?  In theory, Rust packages could use libraries installed in the buildroot and we might need to distinguish them from source code files of the package being scanned.

scripts/CMakeLists.txt
```diff
@@ -36,6 +36,8 @@ if(ENABLE_CSMOCK)
         ${CMAKE_CURRENT_SOURCE_DIR}/run-bandit.sh
         ${CMAKE_CURRENT_SOURCE_DIR}/run-pylint.sh
         ${CMAKE_CURRENT_SOURCE_DIR}/run-shellcheck.sh
+        ${CMAKE_CURRENT_SOURCE_DIR}/convert-clippy.py
+        ${CMAKE_CURRENT_SOURCE_DIR}/clippy.sh
```
kdudka, line 40: We maintain the list of files to install alphabetically sorted, too.

kdudka (changes requested):
I was able to get the results on my first attempt, which is good!  A few remarks inline...

scripts/clippy.sh
petr-matousek: this could be non-trivial to handle correctly and I'd rather wait to see any real-life situation like that. I've added FIXME for now.

petr-matousek:
@kdudka hi Kamil! sorry for late follow up, I was on PTO. thank you for your review. I believe I've addressed your comments, please have a look when you have time. thanks.

scripts/convert-clippy.py
```diff
@@ -0,0 +1,51 @@
...
+PACKAGE_ID_PATTERN  = r"path\+file://(.*)\)"
+# eg -- "package_id":"path+file:///builddir/build/BUILD/stratisd-3.6.5#stratisd@3.6.5"
+PACKAGE_ID_PATTERN2 = r"^path\+file:\/\/([^#]+)(?:#([^@]*))?(?:@(.+))?$"
+
+def main():
+    for line in sys.stdin:
+        try:
+            item = json.loads(line)
+        except:
+            continue
```
kdudka, line 19: Please provide a diagnostic message on `stderr` if a line fails to be parsed.

scripts/inject-clippy.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+ORIGINAL_BINARY="/usr/bin/cargo"
```
kdudka, line 3: I would rename `ORIGINAL_BINARY` to `ORIGINAL_LOCATION` because after invoking the `mv` command, it does not refer to the original binary any more.

kdudka (changes requested):
@petr-matousek Thanks for the update!  A few minor remarks inline.  Are you going to squash the fix-up commits or should I do it on your behalf?

petr-matousek:
@kdudka hi Kamil, all suggestions integrated. thanks for the review!

kdudka (commented):
@petr-matousek Thanks for the update!  I have squashed the existing fixup commits and added one more: da474ec0830789dee05dc5fcb4814fd755623dc7  I think this is ready for merge (with the last fixup commit squashed).  Could you please double check?

petr-matousek:
@kdudka lgtm, thanks!

scripts/convert-clippy.py
```diff
@@ -0,0 +1,53 @@
...
+            sys.stderr.write(f"rust-clippy: Error while converting results: {e}\n")
+            sys.stderr.flush()
+            continue
+
+        if item["reason"] != "compiler-message":
+            continue
+
+        if (not (match := re.search(PACKAGE_ID_PATTERN, item["package_id"])) and
+            not (match := re.search(PACKAGE_ID_PATTERN2, item["package_id"]))):
+            continue
```
lzaoral, line 28: Note that the walrus operator (`:=`) is NOT supported by Python 3.6!

lzaoral: I'd guess that this is also the reason why the following SPEC hack was included in this PR:
```
# needed for /usr/share/csmock/scripts/convert-clippy.py
%global _python_bytecompile_errors_terminate_build 0
```

lzaoral: @kdudka: I'll re-review this PR next week.

kdudka: @lzaoral The hack (suggested by me) is there to fix byte-compilation on el7.  There is no problem with byte-compilation on el8, which uses Python 3.6.

kdudka: ... and the byte-compilation on el7 was failing because `rpmbuild` defaults to Python 2.7 for Python scripts in unversioned directories.

lzaoral: Thank you for the explanation, it should be included in the specfile as well to prevent confusion.

lzaoral (commented):
As already mentioned in comment above, the scanning will fail, if `csmock` is executed on RHEL 8 host or older:
```python
>>> 2024-04-29 12:10:09	"/usr/share/csmock/scripts/convert-clippy.py < /tmp/csmockybgzgpm7/keylime-agent-rust-0.2.4-1.fc40/debug/raw-results/builddir/clippy-output.txt | csgrep --remove-duplicates > /tmp/csmockybgzgpm7/keylime-agent-rust-0.2.4-1.fc40/debug/uni-results/clippy-capture.err"
  File "/usr/share/csmock/scripts/convert-clippy.py", line 26
    if (not (match := re.search(PACKAGE_ID_PATTERN, item["package_id"])) and
                   ^
SyntaxError: invalid syntax

scan.ini: time-finished = 2024-04-29 12:10:09
scan.ini: exit-code = 0
```

@petr-matousek, please, fix these issues in a follow-up PR.

scripts/convert-clippy.py
lzaoral: Thank you for the explanation, @kdudka.  @petr-matousek could add a comment with explanation in the follow-up PR, please?

scripts/inject-clippy.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+ORIGINAL_LOCATION="/usr/bin/cargo"
+NEW_LOCATION="/usr/bin/cargo_original"
+
+if [ -f "$NEW_LOCATION" ]; then
+    rm "/builddir/clippy-output.txt"
+    exit 0
+fi
```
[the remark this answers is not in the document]
kdudka: The specific file is created by the script though.  If the original cargo binary has been moved, it would be a bad idea to try to move it again in case the script is invoked repeatedly on a single chroot (e.g. with `csmock --skip-init ...`).

PR #159, opened 2024-04-22
results: set the "imp" flag in the list of important findings

kdudka:
... rather than clearing it.  If scan results are processed later on, it is useful to know the original state of the "imp" flag, regardless of the context (whether we have a list of important findings only, or an all-in-one list of findings).  A side effect will be that a red `[important]` tag will appear in the HTML output next to each finding, which is probably harmless.

Related: https://issues.redhat.com/browse/OSH-343
Related: https://issues.redhat.com/browse/OSH-565

### DIFF
--- a/py/common/results.py
+++ b/py/common/results.py
@@ -314,7 +314,7 @@ def finalize_results(js_file, results, props):
                     % (js_file, chk_re, csgrep_args)
 
         # finally take all defects that were tagged important by the scanner already
-        cmd += " | csgrep --mode=json --set-imp-level=0 --remove-duplicates"
+        cmd += " | csgrep --mode=json --set-imp-level=1 --remove-duplicates"
         cmd += f" <(csgrep --mode=json --imp-level=1 '{js_file}') -"
 
         # write the result into *-imp.js


py/common/results.py
```diff
@@ -314,7 +314,7 @@ def finalize_results(js_file, results, props):
                     % (js_file, chk_re, csgrep_args)
 
         # finally take all defects that were tagged important by the scanner already
-        cmd += " | csgrep --mode=json --set-imp-level=0 --remove-duplicates"
+        cmd += " | csgrep --mode=json --set-imp-level=1 --remove-duplicates"
```
hanchuntao, line 317: From the command: https://issues.redhat.com/browse/OSH-565?focusedId=24571174&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-24571174, I don't know if the change can work

kdudka: This command uses `--mode=json` whereas the comment you mention is about `--mode=sarif` and the limitation has been removed meanwhile via https://github.com/csutils/csdiff/pull/173.

hanchuntao (approved):

PR #165, opened 2024-05-17
snyk: explicitly enable networking for snyk client execution

lzaoral:
... in the chroot.  Otherwise, it is going to fail.

Related: https://issues.redhat.com/browse/OSH-69


lzaoral:
@kdudka The test-suite failure is unrelated to this PR and is caused by `dnf5` being the default in Rawhide.

PR #172, opened 2024-06-12
csmock: use `Path.name` to get the basename

lzaoral:
... and not `Path.stem` which also removes a filename suffix.

Fixes: b2c720159d52ac89e10e290237f230dbe41848a8 ("csmock: use the root mock profile property for lock names")
Resolves: https://issues.redhat.com/browse/OSH-464


kdudka (approved):
Looks good.  Thanks!

PR #174, opened 2024-06-19
clippy: do not fail the scan if clippy fails to install

kdudka:
If the `clippy` package is not available in the build repos, a warning
is emitted and `clippy` is not recorded as an enabled tool in the scan
properties.  But the scan continues without `clippy` in this case.

The approach is similar to what `--gcc-analyze` does.  This is needed
because `clippy` is not available for each mock config and we do not
want to configure the set of enabled tools separately for each mock
config.

Related: https://issues.redhat.com/browse/OSH-30

### DIFF
--- a/py/plugins/clippy.py
+++ b/py/plugins/clippy.py
@@ -1,6 +1,6 @@
 import os
 
-import csmock.common.util
+from csmock.common.util import write_toolver_from_rpmlist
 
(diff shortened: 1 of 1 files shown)


lzaoral (approved):

kdudka:
@lzaoral Thanks for review!

PR #183, opened 2024-08-16
results: fix quoting in `handle_known_fp_list()`

kdudka:
... by using `shlex.quote()` from standard Python library.  This bug caused regular expressions with parentheses to be interpreted by shell:
```
/bin/sh: -c: line 1: syntax error near unexpected token `('
```

Also avoid using the open-coded `shell_quote()` function from `util.py` while printing shell commands in `exec_cmd()` because the function is rather problematic and should be eventually unimplemented.

Unfortunately, we cannot easily nest `shlex.quote()` while combining `mock --chroot ...` with `su -c ...` and the like because it results in totally unreadable `scan.log`.

Resolves: https://issues.redhat.com/browse/OSH-617

### DIFF
--- a/py/common/results.py
+++ b/py/common/results.py
@@ -21,6 +21,7 @@ import datetime
 import errno
 import os
 import re
+import shlex
 import shutil
 import signal
 import socket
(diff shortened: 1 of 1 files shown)


rhyw (approved):
I also scheduled a task on stage, although the task is still failed, the error message in question is gone.

LGTM

kdudka:
@rhyw Thank you for the review and testing!

PR #184, opened 2024-09-06
Use the term "findings"

siteshwar:
... instead of "defects" in the downstream patches reports.

Related: https://github.com/openscanhub/openscanhub/issues/297


siteshwar:
It should be fixed in reports like https://openscanhub.fedoraproject.org/task/9484/log/coreutils-9.5-9.fc41/scan-results.html

kdudka (approved):

Issue #191, opened 2024-11-07
`csmock` fails to find shellscripts


siteshwar:
This seems to be a regression introduced by https://github.com/csutils/csmock/commit/136730c92f6e214dabfcd8a3b23eaf60d7ca7383.

OpenScanHub scans report below error in [stdout.log](https://openscanhub.fedoraproject.org/task/21349/log/stdout.log):

```
>>> 2024-11-06 16:21:22	"/usr/bin/mock" "-r" "fedora-rawhide-x86_64" "--plugin-option=tmpfs:keep_mounted=True" "--config-opts=print_main_output=True" "--quiet" "--chroot" "SC_RESULTS_DIR=/builddir/shellcheck-results SC_BATCH=1 SC_TIMEOUT=30 /usr/share/csmock/scripts/run-shellcheck.sh /builddir/build/BUILDROOT/*"
Looking for shell scripts...find: '/builddir/build/BUILDROOT/*': No such file or directory
 done
```

PR #192, opened 2024-11-07
csmock: support buildroot location used by new `rpm`

kdudka:
The `%{buildroot}` directory we were looking into is no longer used by new versions of `rpm`.  The directory was moved with the following backward-incompatible change: https://github.com/rpm-software-management/rpm/commit/9d35c8df497534e1fbd806a4dc78802bcf35d7cb

This commit makes csmock look into both the locations.

Fixes: https://github.com/csutils/csmock/issues/191


siteshwar (approved):
Approving, although I have not tried this change in the staging environment.

kdudka:
@siteshwar Thanks for review!

PR #198, opened 2025-04-04
pylint: modernize the `run-pylint.sh` script

kdudka:
... to make it work the same way as `run-shellcheck.sh` works now.

Given the fact that `pylint` is not enabled by default, the `run-pylint.sh` script has not been actively maintained.  This commit transfers the well-tested parts of code from `run-shellcheck.sh` to `run-pylint.sh`.

### DIFF
--- a/scripts/run-pylint.sh
+++ b/scripts/run-pylint.sh
@@ -1,11 +1,41 @@
-#!/bin/sh
+#!/bin/bash
+# shellcheck disable=SC2317
+
 export LC_ALL="C"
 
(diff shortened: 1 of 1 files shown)


jamacku (approved):
LGTM 👍

kdudka:
@jamacku Thanks for review!  I have reported the `SC2317` false positives as https://github.com/koalaman/shellcheck/issues/3183.

siteshwar (approved):

kdudka:
Thank you both for review!

PR #199, opened 2025-04-04
ci(shellcheck): use `sarif-fmt` as display engine

jamacku:
Also, upload the debugging SARIF artifact only when debug is enabled.

Example of difference between ShellCheck display engines: https://github.com/redhat-plumbers-in-action/differential-shellcheck?tab=readme-ov-file#display-engine


kdudka (approved):
Looks good, thanks!

PR #200, opened 2025-04-22
csmock: revert differences in source file paths

kdudka:
... introduced by newer versions of RPM in Fedora.  This is needed for previously recorded known-false-positives to work because they would not be matched with file paths like this (see the duplicated version string in the first two directory levels):
```
util-linux-2.42.start-build/util-linux-2.42.start/disk-utils/cfdisk.c:875:28: warning[-Wanalyzer-malloc-leak]: leak of ‘*m.ignore’
```


siteshwar (approved):

kdudka:
@siteshwar Thanks for review!

PR #205, opened 2025-06-26
make SC_JOBS configurable

sfowl:
This is so users can set their own value for SC_JOBS. Relevant in environments like k8s, where `getconf _NPROCESSORS_ONLN` will report the available number of cpu cores available on host machine, not the container.


lzaoral:
Thank you, @sfowl!  BTW, if you want to get number of CPUs available to given process (as opposed to the number of online CPUs), you can set `SC_JOBS` to `$(nproc)`.  See https://github.com/coreutils/gnulib/blob/84f19f053c483ae5268c580c9629abc2d08d41f5/lib/nproc.c#L247-L255 for additional details.

sfowl:
> Thank you, @sfowl! BTW, if you want to get number of CPUs available to given process (as opposed to the number of online CPUs), you can set `SC_JOBS` to `$(nproc)`. See https://github.com/coreutils/gnulib/blob/84f19f053c483ae5268c580c9629abc2d08d41f5/lib/nproc.c#L247-L255 for additional details.

Interesting! I didn't know about that method, although testing on openshift it doesn't seem to be cgroup aware (as in, it reports 16 cores from inside a container which has a cpu resource request+limit of 1). I think we can discuss alternate detection methods elsewhere though, for now just having this option configurable is a good starting point.

kdudka:
Merging, thanks!

PR #216, opened 2025-11-10
`csmock --embed-context`: use `csgrep-static` in chroot

kdudka:
... so that `csdiff` does not need to be available in build repos

Related: https://github.com/csutils/csdiff/pull/196
Related: https://github.com/csutils/csdiff/pull/239

### DIFF
--- a/csmock/csmock
+++ b/csmock/csmock
@@ -49,6 +49,9 @@ CSMOCK_DATADIR = "/usr/share/csmock"
 
 CWE_MAP_FILE = CSMOCK_DATADIR + "/cwe-map.csv"
 
+# path to the csgrep-static executable
+CSGREP_STATIC = "/usr/libexec/csgrep-static"
+
 CSMOCK_SCRIPTS = CSMOCK_DATADIR + "/scripts"
 
 CHROOT_FIXUPS = CSMOCK_SCRIPTS + "/chroot-fixups"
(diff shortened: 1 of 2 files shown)


kdudka:
I did not realize that `csdiff-static` is not available on RHEL-9 and older distros.  Marking as draft...

sfowl:
@kdudka Now that https://github.com/csutils/csdiff/pull/239 has merged, I think it would be good to now retry this. Would you like to re-open the PR? Alternatively, I could submit the same change in a new PR

kdudka:
@sfowl Sure, reopening.  Thank you for taking care of the `epel-8` builds of `csdiff-static`!

csmock/csmock
```diff
@@ -1289,7 +1292,7 @@ cd %%s*/ || cd *\n\
                 if props.embed_context > 0:
                     # embed context lines from source program files
                     tmp_file = f"{all_file}.tmp"
-                    csgrep_cmd = f"csgrep --mode=json --embed-context {props.embed_context}"
+                    csgrep_cmd = f"{CSGREP_STATIC} --mode=json --embed-context {props.embed_context}"
```
sfowl, line 1295: Running a command like `grep -r \"csgrep csmock` there are quite a lot of calls to the non-static csgrep, though from grep alone I can't easily tell if these calls are made in the mock buildroot or not.

Is the only call that needs to be updated or are there possibly more?

kdudka: Yes, this is the only csdiff/csgrep command that runs in the chroot.  It runs in the chroot because it reads the source code files using absolute paths that are valid in the chroot.  This is an optional feature.  If `csgrep` or `csgrep-static` fails to run in the chroot, `csmock` fallbacks to use the originally captured SAST scanning results without the embedded source code snippets.

sfowl (approved):

PR #225, opened 2026-05-27
feat(shellcheck): add SC_SKIP_JINJA option to skip Jinja2 templates

sfowl:
ShellCheck cannot parse Jinja2 template syntax ({{ }}, {% %}, {# #}). Any file containing these constructs breaks the parser entirely, producing unreliable findings (SC1054, SC1073, SC1083). Setting SC_SKIP_JINJA=1 causes filter_shell_scripts() to exclude these files before they reach shellcheck. Disabled by default so existing behavior is unchanged.

Tested against https://github.com/openshift/cac-content-fork where findings dropped from 11,776 to 43 (99.6% noise reduction).

Resolves: PSSECAUT-1577

### DIFF
--- a/scripts/run-shellcheck.sh
+++ b/scripts/run-shellcheck.sh
@@ -15,6 +15,9 @@ export SC_JOBS
 test -n "$SC_TIMEOUT" || export SC_TIMEOUT=30
 test 0 -lt "$SC_TIMEOUT" || exit $?
 
+# skip files containing Jinja2 template syntax ({{ }}, {% %}, {# #})
+test -n "$SC_SKIP_JINJA" || export SC_SKIP_JINJA=0
+
 # directory for shellcheck results
 test -n "$SC_RESULTS_DIR" || export SC_RESULTS_DIR="./shellcheck-results"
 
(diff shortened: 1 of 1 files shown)


scripts/run-shellcheck.sh
```diff
@@ -59,6 +62,11 @@ fi
 # implementation of the script that filters shell scripts
 filter_shell_scripts() {
     for i in "$@"; do
+        # skip Jinja2 templates if requested
```
kdudka: This would be a disaster from performance point of view because we would execute a separate `grep` process for each single file in the buildroot (including binaries and non-executable data files), whereas we only need to run `grep` for the initially selected shell scripts.  Please consider squashing my fixup commit 8c2e7595a3043bb6f30b622c1017a797b833cde3.

kdudka (changes requested):

scripts/run-shellcheck.sh
sfowl: Thank you! Much better :) I squashed both extra commits.

kdudka (approved):
Looks good.  Thanks!

